PR #4, opened 2013-02-06
Serverside template

llonchj:
Server side templating proof of concept


jpmckinney:
Restarted in #10 #13 #14 #15

PR #14, opened 2013-04-27
RESTful api

llonchj:


Digenis:
Closing because of merge conflicts and inactivity.
Note: the changeset is included in #14, #15 
while passing a version to the schedule resource is already implemented in #123

jpmckinney:
Noting that to keep things simple for maintainers, we would not add new dependencies like CorePost for API helpers, but would rely only on Twisted (also used by Scrapy, so maintainers only need to learn one framework).

PR #15, opened 2013-04-27
Redesigned Web Interface

llonchj:


Digenis:
Closing because of merge conflicts and inactivity.
Note: This would have probably needed a merge with #16.
Passing a version to the schedule resource is already implemented in #123

jpmckinney:
Spirit continued in #304

PR #30, opened 2013-12-02
Add recipe for a centos rpm package of ScrapyD

boogheta:
Repropose the pull request #29 without the undesired commit and with an improved postinstall test on whether scrapyd is well running.


boogheta:
I still do not really understand the reasons this was not pulled : the two concerns were on the use of upstart which as explained is fine with CentOS 6+, and on the dependencies, which are all adressed when installing scrapyd already, which can be done via pip and says so in the rpm installer. Did you guys make any progress on building a dedicated rpm for Scrapy maybe?

jpmckinney:
Linux-flavor-specific configuration is no longer tracked in this repository. #439

Also, deployment is quite simple without packaging. https://github.com/scrapy/scrapyd/issues/258#issuecomment-1126225622

PR #36, opened 2014-02-25
First attemp on Template and Bootstrap UI support

eLRuLL:
Hi,

I am trying to add WEB UI support for the Scrapyd service.

This is a first attemp, so I think this **shouldn't**  be merge as it is. I want to explain first the things I am trying to do, and also ask what to do next.
- First, I saw that you were using plain string inside the `website.py` to populate the pages on the `scrapyd` service, so I am trying to use `Jinja2` to better use _templates_ (like django) to create separate html files for those pages.
- I added **bootstrap 3.1.1** to a `static/` folder, so we can use that css framework. I think this could also be achieved directly pointing to the `css` an `js` files on an external server, so please let me know what could be the best approach here.
- A folder `templates/` has been created, there should be all the html files for the website, like `home`, `jobs`, `available projects`, etc. Now I have there the _home.jinja_  file, that is intended to show the welcome page when you go to `http://localhost:6800/`

**QUESTION**: Could someone help me on how to upload or add `static` when you install `scrapyd`?
So my intention is that every installation should contain its own `templates/` and `static/`folder. I read that it should be done on the `MANIFEST.in` file, but I am really lost there.

It shouldn't be so difficult, could someone maybe share what lines should go there? or maybe somewhere else?

**EDIT:**
- The static files have been added, and now it has support for css an js files.
- I also added the jinja2 requirement on the setup.py file.

Thank you.


Digenis:
Closing because of merge conflicts.
Don't hesitate to reopen if you are still interested.

jpmckinney:
Spirit continued in #304

PR #57, opened 2014-07-27
Renovate scrapy upstart job a bit

CameronNemo:
No need to redirect stdout/err, upstart will catch that. Use chdir stanza to keep up to date with init script, do not source empty/useless defaults file.


jayzeng:
LGTM

dangra:
LGTM.

PR #63, opened 2014-09-16
If the items_dir option doesn't have a file scheme, a file URI won't be used

boechat107:
This fix the issue 56. Relative path wasn't working for the `items_dir`.


jayzeng:
@boechat107 thanks for the pr, can you resolve the merge conflict?

boechat107:
This issue was solved by another [pull request](https://github.com/scrapy/scrapyd/pull/73). That solution seems to be better.

PR #67, opened 2014-11-08
Fix: FEED_URI was always overriden by scrapyd

jurov:
FEED_URI setting under scrapyd had no effect, it always saved feeds into file - see reports such as:
http://stackoverflow.com/questions/15955723/saving-items-from-scrapyd-to-amazon-s3-using-feed-exporter and https://groups.google.com/forum/#!topic/scrapy-users/pl4_TpkLlFY


jayzeng:
@jurov can you pull the upstream code? It has merge conflict

Issue #28, opened 2013-10-29
Specific log file when launching a spider with scrapyd


m-vdb:
Hi there,

I'm currently launching my spiders with scrapyd, and I have some spiders that I can launch several times with distinct arguments (I call these ones generic spiders). The thing is, I want to have distinct log files according to those arguments, and I can't find anywhere any configuration to do that. So my first question is: is it possible to configure scrapyd to do so ?

Then I took a look at your code, and find out interesting lines where we could do that without being too intrusive, right [here](https://github.com/scrapy/scrapyd/blob/master/scrapyd/environ.py#L38-L39). It might be achieved by simply passing an optional log_file argument in messages. Do you think it could be possible ?

Thanks !

sujaymansingh:
@m-vdb Did you find anything for this? I have a fork that might help:
https://github.com/EDITD/scrapyd/tree/editd-fixes

https://github.com/EDITD/scrapyd/commit/178623ed12a473c7dec63bc62eb8b541bf9d01f7

I also submitted a PR, but it looks like scrapyd is no longer being maintained :( What a pity! But let me know if the fork helps.

m-vdb:
@sujaymansingh that's pretty cool! well we implemented an dirty kludgy in-house solution for that... :(

jayzeng:
@m-vdb Thanks for the PR, I code reviewed, let me know what you think.

m-vdb:
I didn't provide the PR, it's @sujaymansingh :)

jayzeng:
Oops, my bad @m-vdb . @sujaymansingh please take a look at the PR :)

PR #71, opened 2014-12-12
[#28] Control log filename with logs_filename cfg

sujaymansingh:
Note that if logs_filename is used, then the deleting of old log files
will not happen. This is because the log file could be anything.
So we can't know which files 'belong' to which spider, and thus we can't
sort and delete old ones.
Perhaps this is acceptable (anyone who wants to control the log_filename
will have to manage their own log directory).

### DIFF
--- a/scrapyd/environ.py
+++ b/scrapyd/environ.py
@@ -1,3 +1,4 @@
+import datetime
 import os
 from urlparse import urlparse, urlunparse
 
@@ -12,6 +13,7 @@ class Environment(object):
     def __init__(self, config, initenv=os.environ):
         self.dbs_dir = config.get('dbs_dir', 'dbs')
         self.logs_dir = config.get('logs_dir', 'logs')
+        self.logs_filename = config.get('logs_filename', '')
         self.items_dir = config.get('items_dir', 'items')
         self.jobs_to_keep = config.getint('jobs_to_keep', 5)
         if config.cp.has_section('settings'):
@@ -30,10 +32,38 @@ class Environment(object):
         if project in self.settings:
             env['SCRAPY_SETTINGS_MODULE'] = self.settings[project]
         if self.logs_dir:
-            env['SCRAPY_LOG_FILE'] = self._get_file(message, self.logs_dir, 'log')
+            env['SCRAPY_LOG_FILE'] = self._get_log_file(message)
         if self.items_dir:
             env['SCRAPY_FEED_URI'] = self._get_feed_uri(message, 'jl')
         return env
+
+    def _get_log_file(self, message):
+        """Combine the logs_dir and logs_filename config items, and substitute
+        any variables in logs_filename to get the full filename of the log file
+        """
+        if not self.logs_filename:
+            # if no filename specified, fall back on the default.
+            return self._get_file(message, self.logs_dir, 'log')
+
+        now = datetime.datetime.now()
+        format_args = {}
+        format_args['project'] = message['_project']
+        format_args['spider'] = message['_spider']
+        format_args['Y'] = now.year
+        format_args['m'] = now.month
+        format_args['d'] = now.day
+        format_args['H'] = now.hour
+        format_args['M'] = now.minute
+        format_args['S'] = now.second
+        filename = self.logs_filename.format(**format_args)
+
+        full_filename = os.path.join(self.logs_dir, filename)
+
+        containing_dir = os.path.dirname(full_filename)
+        if not os.path.exists(containing_dir):
+            os.makedirs(containing_dir)
+
+        return full_filename
 
     def _get_feed_uri(self, message, ext):
         url = urlparse(self.items_dir)
--- a/scrapyd/environ.py
+++ b/scrapyd/environ.py
@@ -59,9 +59,9 @@ class Environment(object):
 
         full_filename = os.path.join(self.logs_dir, filename)
 
-        containing_dir = os.path.dirname(full_filename)
-        if not os.path.exists(containing_dir):
-            os.makedirs(containing_dir)
+        log_dir = os.path.dirname(full_filename)
+        if not os.path.exists(log_dir):
+            os.makedirs(log_dir)
 
         return full_filename
 
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -68,6 +68,23 @@ storing logs set this option empty, like this::
 
     logs_dir =
 
+logs_filename
+-------------
+
+The filename (appended to ``logs_dir``) to use for the crawl. ``{project}``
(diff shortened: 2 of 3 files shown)


scrapyd/environ.py
```diff
@@ -30,11 +32,39 @@ def get_environment(self, message, slot):
...
+        format_args['d'] = now.day
+        format_args['H'] = now.hour
+        format_args['M'] = now.minute
+        format_args['S'] = now.second
+        filename = self.logs_filename.format(**format_args)
+
+        full_filename = os.path.join(self.logs_dir, filename)
+
+        containing_dir = os.path.dirname(full_filename)
+        if not os.path.exists(containing_dir):
```
jayzeng: do you need to ensure containig_dir is writeable?

scrapyd/environ.py
```diff
@@ -30,11 +32,39 @@ def get_environment(self, message, slot):
...
+        format_args['m'] = now.month
+        format_args['d'] = now.day
+        format_args['H'] = now.hour
+        format_args['M'] = now.minute
+        format_args['S'] = now.second
+        filename = self.logs_filename.format(**format_args)
+
+        full_filename = os.path.join(self.logs_dir, filename)
+
+        containing_dir = os.path.dirname(full_filename)
```
jayzeng: containing_dir sounds odd to me, what about log_dir?

scrapyd/environ.py
sujaymansingh: We could test for it, but what should we do if we can't write to it?

I can't find any relevant Exception to throw. How would want this to be handled?

(In practice, trying to create the dir will raise an Exception if the dir isn't writable and the process dies.)

scrapyd/environ.py
sujaymansingh: Sure, `log_dir` works

scrapyd/environ.py
```diff
@@ -30,11 +32,39 @@ def get_environment(self, message, slot):
...
+        now = datetime.datetime.now()
+        format_args = {}
+        format_args['project'] = message['_project']
+        format_args['spider'] = message['_spider']
+        format_args['Y'] = now.year
+        format_args['m'] = now.month
+        format_args['d'] = now.day
+        format_args['H'] = now.hour
+        format_args['M'] = now.minute
+        format_args['S'] = now.second
```
jayzeng: So the log format becomes:
logs_filename = {spider}-{Y}{m}{d}.log

We should definitely have a default date format but I feel this perhaps should be something configurable, what if users want mm/dd/yyyy for them to integrate into some sort of log parsing solution that they don't have control about specific format?

scrapyd/environ.py
```diff
@@ -30,11 +32,39 @@ def get_environment(self, message, slot):
...
+        filename = self.logs_filename.format(**format_args)
+
+        full_filename = os.path.join(self.logs_dir, filename)
+
+        log_dir = os.path.dirname(full_filename)
+        if not os.path.exists(log_dir):
+            os.makedirs(log_dir)
```
jayzeng: Few things:
- os.makedirs creates a folder with permission 777 (see https://docs.python.org/2/library/os.html#os.makedirs), which we should make 744 or even 400 or 440
- os.makedirs raises an Error exception when the folder already exists or can not be created, so we should catch the exception
- I feel we should make it clear that scrapyd is trying to create a folder at log_dir, let users make the decision whether they want to do any post cleanups.

scrapyd/environ.py
sujaymansingh: I don't quite follow @jayzeng

If the users wanted to use `mm/dd/yyyy` they'd be able to set:

```
[scrapyd]
logs_filename = {m:02d}/{d:02d}/{y:04d}
```

in the scrapy config.

scrapyd/environ.py
sujaymansingh: * Changing the permission is a good idea. I think that `744` is a good default setting.
* Again, we can catch the exception, but what should we do with it?
* There is a note about handling the log_dir when log_filename is set in the documentation
Is there much else we need? https://github.com/sujaymansingh/scrapyd/commit/5cdbb752199c0557e7530e0e3cdb673f3b5efe9b#diff-fd40cf2be7711772de9d8316da038cceR71

jayzeng:
@sujaymansingh sorry for the delay. Yeah, your changes look good to me.

Digenis:
I was trying to fix some race conditions in environ.py
and I found this part really weird.

- First, this doesn't take into account people using feeds.
  `_get_log_file` could also be used for feeds but it's not.
- Someone will have to maintain the format args to keep up with any new parameters.
  #120 is an example and #123 is about to introduce a new one.
- Most important point:
  To treat the logs separately, unnecessary complexity is introduced.
  This could have been resolved by using scrapy's LOG_FILENAME.
  Even if it's too much of a bother, a custom schedule webservice could set it.

I really think we should revert this change to keep the core components as simple as possible.
I'd favour putting such functionality in a webservice in a contrib/ folder.

sujaymansingh:
Hi @Digenis 

I think the key point is whether to support log file naming in scrapyd.

As you mention, maintaining format args is going to have to be done. However, that will be the case no matter what. Either format args are available and maintained, or they shouldn't be available.
(The worst scenario is if they were available and not maintained!)

The reason for this PR was #28. Both myself and @m-vdb had a requirement to control the log file name used by scrapyd. There might be others!

Digenis:
I disagree with the "non-intrusive" part.
It's right in environ.py. conditionally wraping `_get_file`
while not being generic enough to handle both logs and feed exports.
I think it's more complex than it needs to be.

Meanwhile, all it solves is just involving dates in the log directory structure.
(You can search by ctime btw if this solves something)
I have use cases where the logfile needs to be controlled using spider arguments or settings.
Such would fit more #28 which also refers to generic spiders
where one of their configuring kwargs can be a step in a taxonomy.
Eg. you may want this functionality more than date-oriented directory structures.
`curl http://localhost:6800/schedule.json -d project=project_abc -d spider=generic_blog_spider -d blog="pencho's_blog.com" -d with_images=True` to result in such a logfile `project_abc/generic_blog_spider/pencho's_blog.com/905f6a56b.log`

A custom webservice can resolve that (and this PR) without even patching scrapyd.
I think the only reason it didn't happen is because of the lack of ideas.

I don't think we should be hesitant reverting an unreleased feature.
People who use the master branch instead of a stable release
must assume such risks and manage it by closely following any upstream developments.
However I wouldn't revert this without moving the functionality to a webservice first
so everyone who happens to use it does the transition.
It will follow some deprecation cycle only because we are in the unfortunate situation
where people (justly) think that scrapyd is not being released any more
so they just install the master branch from github.

PR #86, opened 2015-03-25
Moved scrapyd-deploy into separate scrapyd-client package

### LINKED ISSUE #85: Add scrapyd-client package
We should add scrapyd-client package for users so they won't need to install the full scrapyd package to deploy their code.

rdowinton:
Resolves #85 

A couple of questions:

Is the Twisted dependency needed? If not then it can be removed.
I noticed the scripts in the current scrapyd package are outside of the package directory ('scrapyd'), should this be the same here? I've added the scrapyd-deploy script the scrapyd-client package directory otherwise it would be empty.

Please let me know if there's anything unnecessary in the setup.py that can be removed.


rdowinton:
We should probably also add scrapyd-client to extras_require in scrapyd's setup.py too.

nyov:
Twisted dependency can be removed, it's not used in here.
`'Environment :: No Input/Output (Daemon)',` line can be removed from setup.py

Otherwise I see no issues, but I haven't tested the script itself (should be the same as before, right?)

Optionally could add support for python3 urllib through six.

pablohoffman:
@rdowinton was going to close this pull request and push the client code to a separate repo:
https://github.com/scrapy/scrapyd-client

PR #89, opened 2015-04-10
Changed scripts to be installed using entry_points

utek:
Changed scripts from bin to be run using console_scripts from entry_points.

It allows to run it easy on either linux or windows sot following problem is no longer a problem:
http://stackoverflow.com/questions/22646323/windows-scrapyd-deploy-is-not-recognized


pablohoffman:
Looks good to me. @rdowinton can you merge and then apply to scrapyd-client too?

PR #93, opened 2015-04-22
Accept override jobid with default of UUID in webservice.py

TheRealJoeLinux:
I've been using Scrapyd in production for months now, and this was a very useful feature for us, so viewing a listing of the jobs online was much more human-friendly, and the jobid's were also used for tracking the generated log messages.


jayzeng:
Cool, thanks @TheRealJoeLinux

Digenis:
This should be documented here
https://github.com/scrapy/scrapyd/blob/master/docs/api.rst#schedulejson

PR #101, opened 2015-07-02
Add contributing to the docs

dfockler:


Digenis:
I agree with all this.
Scrapy has more detailed guidelines and this justifies the link.
You put a separate section about tests and that good because scrapy uses tox.

Merge?

jayzeng:
Thanks @dfockler

PR #106, opened 2015-08-28
check service status /status.json

mariotux:
To check the load status of a service

request -> GET http://localhost:6800/status.json 
response -> { "status": "ok", "running": "0", "pending": "0", "finished": "0", "node_name": "node-name" }


Digenis:
On #5 there was an impression that #14  (eventual successor of #3 )
is going to add a `status.json` that is supposed to show the status of a specific job.

In the [actual PR](https://github.com/llonchj/scrapyd/commit/2b39890d09e8b993309972322132b23e977d8490?diff=unified#diff-a048cbc56b216c593ee09ea9dde49ed6R123) it happens to show the status of a project (versions, spiders etc)

Now to revoke the term "actual", the PR has no activity since 2013
so I can hardy think someone would still have claims over the name of the resource.
But since so far I've seen 3 different things being called "status",
to prevent confusion, it could be better each to have a more descriptive name.
eg. jobstatus, projectstatus, daemonstatus
Doesn't this highlight the need for a restful API?

@llonchj are you still interested on #14 ?
Do you plan reviving it, resolving all the merge conflicts etc?

llonchj:
@Digenis It's been a few years since. I can review that in the future. Feel free to use the feature and create a pull request

mariotux:
I think so, @Digenis has reason. The resource of web service must be more semantic to describe what does.

jayzeng:
I think this endpoint is pretty useful, thanks for the PR!

PR #108, opened 2015-09-15
Add docs for #93

dfockler:
Explains about being able to override Scrapyd's generated UUID using jobid as a param to Schedule


jayzeng:
thanks @dfockler

PR #111, opened 2015-10-06
config option for website root class

Digenis:
Make the Root website class configurable in `scrapyd.conf`

Use case:
Allow overriding the website class
without modifying the twisted application itself.
Although it can be done from the app,
there are several pull requests
demonstrating that people mostly focus on the website
rather than the app.


jayzeng:
Cool, thanks!

PR #119, opened 2015-11-27
Debian / Ubuntu .deb packaging issues

darylounet:
- Updated version
- Scrapyd-deploy cleaning after #91
- .deb compiling issue resolved

### DIFF
--- a/debian/scrapyd.install
+++ b/debian/scrapyd.install
@@ -1,3 +1,1 @@
 debian/000-default etc/scrapyd/conf.d
-debian/tmp/usr/bin/scrapyd
-debian/tmp/usr/lib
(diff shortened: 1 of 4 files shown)


debian/scrapyd.install
```diff
@@ -1,3 +1 @@
 debian/000-default etc/scrapyd/conf.d
-debian/tmp/usr/bin/scrapyd
-debian/tmp/usr/lib
```
Digenis: I'm not familiar with debian packaging.
Can you explain me why these lines are deleted?

jpmckinney:
Debian packaging is no longer tracked in this repository. #439

PR #120, opened 2016-01-15
Forgotten job placeholder in the 'logs_filename' setting

decaz:
Added job placeholder {job} into the 'logs_filename' setting.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -71,16 +71,15 @@ storing logs set this option empty, like this::
 logs_filename
 -------------
 
-The filename (appended to ``logs_dir``) to use for the crawl. ``{project}``
-and ``{spider}`` placeholders will be substituted, as will certain datetime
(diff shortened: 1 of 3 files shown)


Digenis:
Looks good to me.
I'd squash if I were you.

Digenis:
@decaz, can you squash?

decaz:
I have already removed fork.. But I can create a new PR with a single commit - is it OK?

Digenis:
Well my opinion is that we should revert the whole logs_filename feature
in favour of some scrapyd.contrib webservice that does the formatting
because it's too much for environ.py to be occupied with this kind of things.
In this case there's no point in starting over.

decaz:
I agree, though this PR can be merged because it's simply a feature, which subsequently can be taken into account during refactoring you talking about.

Digenis:
Squashed and merged in da66373c8f871dc40fae97174930e94c4aa5670b.

PR #122, opened 2016-01-27
Dynamic requirements for different pythons

Digenis:
Also, it would be interesting to see if just backporting #62 can make it compatible with 1.0.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 from os.path import join, dirname
+import sys
 
 with open(join(dirname(__file__), 'scrapyd/VERSION')) as f:
     version = f.read().strip()
@@ -34,6 +35,10 @@ try:
 except ImportError:
     from distutils.core import setup
 else:
-    setup_args['install_requires'] = ['Twisted>=8.0', 'Scrapy>=0.17']
+    if sys.version_info < (2, 7):
+        setup_args['install_requires'] = ['Twisted>=8.0,<=15.1', 'Scrapy>=0.17,<0.19', 'w3lib<1.9']
+    else:
+        setup_args['install_requires'] = ['Twisted>=8.0', 'Scrapy>=0.17,<0.25', 'w3lib<1.9']
+
 
 setup(**setup_args)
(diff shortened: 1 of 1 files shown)


Digenis:
Also, it would be interesting to see if just backporting #62 can make it compatible with 1.0.

Digenis:
It looks like python2.6 support needs to be dropped to use scrapy-0.24.
Should python2.6 support be dropped on a bugfix version? I think not.
However I'd still consider backporting #62.

Digenis:
I need a comment from the developers on this.
scrapyd-1.0 still supports py2.6 so the only way to set version dependencies properly
is to restrict its scrapy version to 0.18.
Otherwise py2.6 support will be dropped on a bugfix release.

The only workaround I can think of is using version_info in setup.py to decide scrapy's max version.
Does this sound sane?

Digenis:
This PR is ready.
Anyone willing to review it?

redapple:
@Digenis , I'm not sure I follow (excuse my very poor knowledge of scrapyd)

I don't understand your sentences:
"Should python2.6 support be dropped on a bugfix version? I think not.
 "Otherwise py2.6 support will be dropped on a bugfix release."

My preference would be to drop Python 2.6 support in the next minor or major release of scrapyd.
[Scrapy dropped Py2.6 support nearly 2.5 years ago](https://github.com/scrapy/scrapy/blob/master/docs/news.rst#0200-released-2013-11-08)

Python 3 support is also needed and too many Pythons is already a lot of effort for scrapy itself.

Digenis:
I replaced the last commit, the new one properly disables wheel building.
I do this because the scrapy and twisted version requirements depend on the python version
(as seen in 63ff12f1e103f9830df1fd6d4820386f2bf22eaf)

The reason I add a bugfix on this branch is because of users who ended up depending on some scrapyd-1.0 behavior.

I definitely want to bump a new release on which python2.6 support is dropped 
so that development can go on. I hope this is the last bugfix bump on 1.0.

My question to scraping hub is whether its apt repositories will get updated with this
otherwise I should drop them from the documentation.

redapple:
@Digenis , sorry it's taking so long about the apt repos.
It looks like we'll have resources to work on our repos/PPA/bintray offering soon.

Digenis:
Fixed the w3lib dependency and squashed.
I'll wait for comments a few days and merge.

PR #123, opened 2016-02-03
adding version support using a -d _version=145446.. argument

lookfwd:
Adds support for an interim way to schedule specific spider version. The spider version should already exist. For example in order to list available versions do:

    $ curl http://localhost:6800/listversions.json?project=tutorial
    {"status": "ok", "versions": ["1454462379", "1454462626"]}

version `1454462626` will be scheduled by default. To schedule version `1454462379` you can now do (note the `_` prefix of `_version`):

    curl http://localhost:6800/schedule.json -d project=tutorial -d spider=example -d _version=1454462379


Digenis:
If it's that trivial, I agree that we should include such a feature.
It would be useful to schedule 2 different versions of a spider
to compare the output after an update.

However I worry that we are filling the POST parameters with magic prefixes.
I'm ok with this one.
Perhaps we should start considering a new webservice for scheduling
and deprecate the old one.

lookfwd:
I would consider calling it `version` instead of `_version` and add it to the official non-magic parameters but the risk of existing users using `version` as an argument would have to be assessed. Maybe another name like `egg_version` or `spider_version` would be a better solution.

Digenis:
More like "I worry about the risk of `_version` conflicting with users' spider arguments"
which of course sounds unlikely.
It will not happen accidentally, I don't argue over this.
I think that they are a bit of a burden in some scenarios, eg:
Suppose someone wants to expose all of a spider's kwargs to untrusted user input.
Another step for removing those special variables from user input would be required
and this extra step would have to keep up with what scrapyd considers a special variable.
I shouldn't have called them "magic prefixes",
just "special variables" whose names make them less likely to conflict.

This is not however a discussion for this PR.
I'll test and merge this, probably over the weekend.
It adds much functionality with only a few changes :+1:

Digenis:
A line in the docs is needed before merging.
https://github.com/scrapy/scrapyd/blob/master/docs/api.rst#schedulejson

szhem:
Hello guys,

Is it possible to obtains the spiders from the provided version?
I'm trying to issue the following request that returns spiders only from the latest version

    curl 'http://localhost:6800/listspiders.json?project=myproject&_version=1_0_0'

lookfwd:
Good point, for consistency and completeness of the API. Done

szhem:
@lookfwd thanks a lot for your efforts, unfortunately this issue seems to be not so simple as it seemed to be.

Let's say I have version 1_0_0 with spiders spider_\[1-3] and version 1_0_1 with spiders spider_\[2-3] (spider_1 is deleted), then the following request that schedules spider_1 from version 1_0_0 fails

    $ curl http://localhost:6800/schedule.json -d project=my_project -d spider=spider_1 -d _version=1_0_0
    {"status": "error", "message": "spider 'spider_1' not found"}

... `addversion` also seems to return inconsistent results if adding versions in the opposite order like this:

Let's add version 1.0.1 with spiders spider_\[2-3]

    $ curl http://localhost:6800/addversion.json -F project=my_project -F version=1.0.1 -F egg=@my_spiders-1.0.1-py2.7.egg
    {"status": "ok", "project": "my_project", "version": "1.0.1", "spiders": 2, "node_name": "localhost"}

The number of spiders returned is 2 that is correct

Now let's add version 1.0.0 with spiders_\[1-3] (with new spider - spider_1)

    $ curl http://localhost:6800/addversion.json -F project=my_project -F version=1.0.0 -F egg=@my_spiders-1.0.0-py2.7.egg
    {"status": "ok", "project": "my_project", "version": "1.0.0", "spiders": 2, "node_name": "locahost"}

The number of spiders returned is 2 that is incorrect and can be verified with `listspiders` request

    $ curl 'http://localhost:6800/listspiders.json?project=my_project&_version=1_0_0'
    {"status": "ok", "spiders": ["spider_1", "spider_2", "spider_3"], "node_name": "localhost"}

lookfwd:
@szhem - I will check in a bit. If you have a look to the'eggs' folder you should see files for every version there, right? When you give custom versions, as you do, the sorting order changes. Maybe you would like to form version numbers in such a way that it does sort properly. In any case I will have a more thorough look in a few hours

szhem:
@lookfwd 

> If you have a look to the'eggs' folder you should see files for every version there, right

Yep the files are there. 

> Maybe you would like to form version numbers in such a way that it does sort properly

I suppose that the issue is not in the files ordering, but in versions upload ordering. 

Let's say we have versions 1_0_0 and 1_0_1. The natural order for this versions is (1_0_0, 1_0_1).
But in the samples my upload order was (1_0_1, 1_0_0) and scrapyd returned spiders not from the just uploaded file (although it knows its name and version), but from the last one in natural order from the eggs folder.

lookfwd:
@szhem - I perfectly see your point and I'm glad there's an end-user who can check the code so quickly. It really helps. So indeed, there are two more cases, one regarding schedule and one regarding the addversion you just pointed out where the functionality wasn't correct. I believe those are Bug Fixes, since they wouldn't work in previous versions of scrapyd as well. What I can give you (see last commit), is something that is consistent _BUT_ I don't think it would be wise to change the way things work right now in terms of natural/upload order. More specifically the order is [distutils LooseVersion](http://epydoc.sourceforge.net/stdlib/distutils.version.LooseVersion-class.html) and I would also guess that there are already users that rely on that behaviour. Here's what the latest commit gives:

    # upload version 1_0_1 with two spiders
    ...
    Packing version 1_0_1
    {"status": "ok", "project": "tutorial", "version": "1_0_1", "spiders": 2}

    $ curl 'http://localhost:6800/listspiders.json?project=tutorial'
    {"status": "ok", "spiders": ["example", "example2"]}

    # upload version 1_0_0 with three spiders
    ...
    upload version 1_0_0
    {"status": "ok", "project": "tutorial", "version": "1_0_0", "spiders": 3}

    $ curl 'http://localhost:6800/listspiders.json?project=tutorial'
    {"status": "ok", "spiders": ["example", "example2"]}

    $ curl 'http://localhost:6800/listspiders.json?project=tutorial&_version=1_0_0'
    {"status": "ok", "spiders": ["example", "example2", "example3"]}

Digenis:
Such backward incompatible behaviour changes,
even when considered bugfixes by some,
should indeed be discussed on individual tickets.
If accepted, they are released in a later version
preceded by a deprecation period with warnings and stuff.

I think what @szhem is after is a way to select a default version for scheduling.
There are workarounds.
scrapyd-deploy supports deploying to multiple targets, you can set up a test instance of scrapyd.
Perhaps you can duplicate a deploy target and change its project name, though I haven't tried this.

Now back to this PR:
It needs documentation, (just a line really https://github.com/scrapy/scrapyd/pull/123#issuecomment-180406308)
and some squashing.
I don't understand how UtilsCache fitted into this.

lookfwd:
@Digenis `get_spider_list()` is by default slow because it needs to decompress the egg in a temp directory, spawn another process, and do a `scrapy list` to get the list of spiders. `UtilsCache` is used to do this only once and cache the results. The key for the cache was the project name e.g. `testproj`. As soon as we add version support to `get_spider_list()`, we need to cache the results per-version i.e. have a composite key, now, in the form of e.g. `testproj:0.0.1`. The default project has the key e.g. `testproj:`. The cache should be invalidated when a new version is added, or a version of the project is deleted, or (of course) when a project gets deleted. Out of these cases, only the first one was supported... on all the other cases (deletes) there was no invalidation which was a bug. If you had version 1 and version 2 of a spider and deleted version 2, the API would still keep returning the values from version 1. Anyway, all those are fixed now. It could be done a little bit more efficiently but I think this is good enough (reasonably) assuming that those APIs aren't hit that heavily.

lookfwd:
doc + squash done

scrapyd/webservice.py
```diff
@@ -82,7 +83,7 @@ def render_POST(self, txrequest):
         version = txrequest.args['version'][0]
         eggf = StringIO(txrequest.args['egg'][0])
         self.root.eggstorage.put(eggf, project, version)
-        spiders = get_spider_list(project)
+        spiders = get_spider_list(project, version=version)
```
lookfwd: This was the bug that @szhem mentioned. If I add a spider while a newer project version already exists on the system, the results `get_spider_list` returns were wrong (representing the latest version).

scrapyd/webservice.py
```diff
@@ -140,4 +143,5 @@ def render_POST(self, txrequest):
         project = txrequest.args['project'][0]
         version = txrequest.args['version'][0]
         self._delete_version(project, version)
+        UtilsCache.invalid_cache(project)
```
lookfwd: This is a bit too broad. It could potentially invalidate only 1 or 2 cache entries but I don't think that performance justifies the extra code complexity.

scrapyd/webservice.py
```diff
@@ -82,7 +83,7 @@ def render_POST(self, txrequest):
         version = txrequest.args['version'][0]
         eggf = StringIO(txrequest.args['egg'][0])
         self.root.eggstorage.put(eggf, project, version)
-        spiders = get_spider_list(project)
+        spiders = get_spider_list(project, version=version)
         self.root.update_projects()
         UtilsCache.invalid_cache(project)
```
lookfwd: This is a bit too broad. It could potentially invalidate only 1 or 2 cache entries (instead of every entry for this project) but I don't think that performance justifies the extra code complexity.

szhem:
Hello guys, this is amazing to have fully fledged version support in scrapyd! Now everything works as expected. @lookfwd thanks a lot!

lookfwd:
done

PR #125, opened 2016-02-22
Indicate that wheels are only python 2.7 compatible

Ivoz:
See https://wheel.readthedocs.org/en/latest/#defining-the-python-version

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+python-tag=py27


Digenis:
To my understanding (I'm not very familiar with wheels)
this will should not prevent python2.6 users from installing
but only from building a python2.6 wheel.

Ivoz:
It will prevent 2.6 users from installing a wheel built from this source.

The counter-part in #126 will stop a 2.6 user installing directly from source.

Digenis:
In this case I merge+backport this for the 2.6 compatible branch and merge the other one for the upcoming py27-only release.

Ivoz:
@Digenis friendly FYI, if you would instead like to indicate a source *is* compatible with 2.6 and 2.7, then the `python-tag` to use is `py26.py27`. Or if you are not worried about the existence of <2.6 users, you can simply use `py2`.

Digenis:
I am interested for py2.6 users, but on an older branch.
Actually I figured out how to disable wheels on 2c4c6401025696714ea17cdbdd7c38991dd4559e.
Users of the older branch will have to settle with a non-wheel installation.

This PR is is still acceptable however because py2.6 support will be dropped on the next release.

PR #130, opened 2016-03-10
daemonstatus.json miscounts running & finished jobs

BruceDone:
Hi all : 
   since the daemonstatus.json did not show right count of the current server running count ,then i fix it and confirm it on my machine ,then it is right now . thanks

### DIFF
--- a/scrapyd/webservice.py
+++ b/scrapyd/webservice.py
@@ -25,16 +25,9 @@ class WsResource(JsonResource):
 class DaemonStatus(WsResource):
 
     def render_GET(self, txrequest):
-        running = 0
-        pending = 0
-        finished = 0
-        projects = self.root.scheduler.list_projects()
-        for project in projects:
-            spiders = self.root.launcher.processes.values()
-            running += len(spiders)
-            queue = self.root.poller.queues[project]
-            pending += queue.count()
-            finished += len(self.root.launcher.finished)
+        pending = sum(q.count() for q in self.root.poller.queues.values())
+        running = len(self.root.launcher.processes)
+        finished = len(self.root.launcher.finished)
 
         return {"node_name": self.root.nodename, "status":"ok", "pending": pending, "running": running, "finished": finished}
 


Digenis:
You are right about the bug; there's additional counting for every project.

However there are problems in the commits:

* You prefixed your name in every commit message.
* You didn't complete your sentence in the first commit:
* You seem to have used some misconfigured formatting tool on the code.
  * The empty line after every class declaration is deleted.
  * The empty lines between some imports are deleted.
  * Some wrapping of an irrelevant long line was introduced
    and it's still longer than 80 columns.
* The second commit message is misleading,
  it reads like you update a dependency or a submodule.
  We would conventionally call it `Bump 1.1.1`
* Bumping the version doesn't make sense in this PR
  because the daemonstatus.json webservice was never released.
  In addition, the commits that accumulated since the last bump
  don't qualify for a "bugfix" version.

I suggest starting over. Maybe you need to reconfigure your text editor first.
This should be a single commit and the diff should spawn less than 20 lines. 

Also,
you seem to have forgotten to configure an email in your git config.
This wouldn't bother the project accepting PRs. It's just a reminder.

BruceDone:
thanks for reply ,i will update the commit and code as the your request ,then i will push it .

Digenis:
You are probably not familiar with git.
The commit message is larger than 50 characters.
You said "update the commit" but you actually added a new one.
Also, the formatting changes looks like they were reverted by hand
so even if I instruct you how to squash the 3 commits,
strange diffs will so up in the result.

Read documentation about `git commit --amend`, `git add -p` and `git rebase -i`.
If you are not willing to go through this I'll patch it.
This is not some urgent bugfix since it's on an unreleased version,
so take your time if you see this as an educational experience.

BruceDone:
thanks for your patience and i will fix my commit as the right formate tomorrow

Digenis:
Sorry for the delay.
Github doesn't notify me where there is commit activity on a PR.
I'm testing it right now and will merge today.

Digenis:
perfect :+1:

PR #138, opened 2016-03-16
Change script to enable scrapyd run on Windows

luc4smoreira:
A very simple code change to enable run Scrapyd on Windows

### DIFF
--- a/scrapyd/scripts/scrapyd_run.py
+++ b/scrapyd/scripts/scrapyd_run.py
@@ -9,3 +9,6 @@ import scrapyd
 def main():
     argv[1:1] = ['-n', '-y', join(dirname(scrapyd.__file__), 'txapp.py')]
     run()
+
+if __name__ == '__main__':
+    main()


Digenis:
This doesn't include instructions on how to install on windows.
How did you install? Do you use distutils?

The scrapyd_run.py patch shouldn't be necessary,
setuptools will use the entry_points to create an exe file in windows.

These were some of my comments but not all.


The current status of scrapyd on windows is "Use at your own risk"
so I don't think we should update the documentation before thorough testing.

luc4smoreira:
It´s just to run, not to install as a service.

If you run the script on windows it will start the scrapyd, doing the same as here:

https://groups.google.com/forum/#!topic/scrapy-users/dinsVKGaMUQ

Digenis:
Install as a service can only mean (eventually) to install as a system package including init scripts etc.
Did you just download the code from github?

luc4smoreira:
Sorry, I did not explain right.

Yes. I did the download on my Windows instance from Github and I tried to find a way to run. 
So, this was the solution. Maybe could be useful for other users.

Now I can run on Eclipse with Pydev and debug the code.

Digenis:
You got the master branch from github which is under development.
Try some tool like pip to install packages from pypi.

I'm not against this PR.
I only don't want the doc to give the false impression that scrapyd support windows.

Digenis:
@luc4smoreira, if you want it I'll merge the `scrapyd_run.py` patch
Just configure your editor while you amend the commit
to always leave a trailing newline.
Also, do this https://www.python.org/dev/peps/pep-0008/#blank-lines
(Mostly to keep git-blame clear)

luc4smoreira:
Have a look. Now it's just the change on script to run scrapyd

Digenis:
Just one last thing.
Squash them in a single commit.
e.g. `git reset --soft 1b0b50f995a89236e4aa16db5aeb9cff123196c2^ && git commit`

PR #139, opened 2016-04-06
Limit the maximum jobs per project

luc4smoreira:
Hi.

I developed this new feature to allow limit the maximum jobs per project. Please, check if it is interesting.


jpmckinney:
Replaced by #140

PR #157, opened 2016-07-21
Run tests with tox on Travis CI

redapple:


redapple:
@Digenis , what do you think?

Digenis:
It looks fancy but since travis runs it without problems for now we should be ok.
Having no one familiar with it, it's going to be a trial and error process.

redapple:
Maybe it's just me, or how scrapy does it.
Scrapy has been using Travis CI this way for some time now (https://github.com/scrapy/scrapy/commit/fabb351097a028c3517040529a654e34f7269db3#diff-354f30a63fb0907d4ad57269548329e3)
I tend to use tox all the time, and having the same command running on Travis seems better to me.

Digenis:
Since scrapy has been using it this way for this long, I doubt it'll break now.
:+1:

.travis.yml
```diff
@@ -1,34 +1,42 @@
 language: python
+python: 2.7
```
kmike: I think to enable Python 3.5 it should be python: 3.5

Digenis: There's a separate branch for the porting to python3

Issue #144, opened 2016-05-12
Assigning job priorty while scheduling


DharmeshPandav:
I am having a use case scenario where we are running new job for each task on same domain
I am handling this using `scrapyd` and `schedule.json` API
like 
```
curl http://localhost:6800/schedule.json -d project=myproject -d spider=somespider -d page_url=some_spdier_start_url
```

but there are some cases where, new jobs request needs to be served immediately ( soon after current jobs finishes) [i am using single process  per domain so have only one process running for any spider at any given time]

can we use priority queuing in scrapyd while scheduling job ?
https://github.com/scrapy/queuelib/blob/master/queuelib/pqueue.py
```
#something like
 curl http://localhost:6800/schedule.json -d project=myproject -d spider=somespider -d priority=1
#priority : {1- highest, 2 - high, 3 - normal, 4 - low, 5 - lowest } , default to 3 - normal
```

i noticed that this is implemented on Scrapinhub dashboard when we run a spider

DharmeshPandav:
After looking into code-base and few trial-n-error ,
I see that priority option is already implemented but argument was not documented

it is implemented in `SqlitePriorityQueue` class in `sqlite.py` package and priority argument value was set to `0.0` by default in `add` method of `SqliteSpiderQueue` class in `spiderqueue.py` package

```python
def add(self, name, **spider_args):
        d = spider_args.copy()
        d['name'] = name
        priority = float(d.pop('priority', 0))
        self.q.put(d, priority)
```
i think this argument should be included in documentation 

should i close this issue and open other for documentation fix ?

Digenis:
Indeed, it is missing documentation and tests.
Also, I'd rather move it to the schedule.json webservice
and have it next to the other statements that modify the dictionary
by popping "reserved" arguments.

Digenis:
For 1.1 branch: https://github.com/Digenis/scrapyd/commit/5d068b36924a9df7ec48935407c89b62c4d7394b
I'll commit this one so please comment.

---

For the next release I plan something like https://github.com/Digenis/scrapyd/commit/448284c20bc1b4326d51a7b7489d13e8886ad952
I'll go through a PR for it because it's backwards incompatible
and I actually want to change it in an even more backwards incompatible way
because passing the spider arguments with `**spider_args`
is an abuse of the keyword arguments feature
which leads to colliding argument names
(eg try `(lambda a, **k: 0)(1, **{'a':2})`)

DharmeshPandav:
[...]

PR #161, opened 2016-07-22
Move priority setting to schedule.json

Digenis:
Continued from https://github.com/scrapy/scrapyd/issues/144#issuecomment-218963899

The arguments passed from component to component contain arguments addressed to different components, arguments addressed to the crawl process all in a single dictionary. The components shouldn't be poping keys from the dictionary containing the crawl arguments.

Priorities are of interest to schedulers and queues, not processes.
That's why I make the setting a positional argument in their method calls.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -67,6 +67,7 @@ Schedule a spider run (also known as a job), returning the job id.
   * ``spider`` (string, required) - the spider name
   * ``setting`` (string, optional) - a Scrapy setting to use when running the spider
   * ``jobid`` (string, optional) - a job id used to identify the job, overrides the default generated UUID
+  * ``priority`` (float, optional) - priority for this project's spider queue — 0 by default
   * ``_version`` (string, optional) - the version of the project to use
   * any other parameter is passed as spider argument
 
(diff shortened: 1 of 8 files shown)


Digenis:
Needs testing with custom ws, scheduler and queue.
Also, changelog.

Digenis:
TODO (later): document testing custom components with zope.interface.verify

Digenis:
Will merge if no comment.
Review welcome.

my8100:
I have never use the *priority* option before. 
Would try and test it this weekend.

Digenis:
It's meant to set the priority of a scheduled run
but only within the project's queue
which makes it useless in most multi-project cases.
Especially when the pending queue never empties
#187 becomes a showstopper.

my8100:
It seems that jobs of project 'TESTING2' would not be scheduled
until there are no pending jobs of project 'TESTING' due to #187,
no matter how high the priority of project 'TESTING2' is.

BTW, should/could we show the priority info in the jobs page,
as well as the API `listjobs.json`?

Digenis:
Makes sense, yes.

Digenis:
any comment before merge?

my8100:
No, I think it’s ready to go.

my8100:
> BTW, should/could we show the priority info in the jobs page,
> as well as the API `listjobs.json`?

Implemented in PR #343.

PR #163, opened 2016-07-23
Deprecate alternative sqlite backed queues/dictionaries

Digenis:
_Work in progress because it will conflict with #151_ 

There are 3 alternatives sqlite backed queues and 3 more dictionaries.
I don't think we need so many alternatives in internal components.
In this case I think they are just leftovers from when scrapyd separated from scrapy.

The easiest way to reduce them would be to keep the json variations,
which are used by default, and deprecate the rest.

I have a preference fo the pickling ones because it's easier to handle dates and other objects
but 2 arguments against it convince me to go with the json ones:
- It would make the database more cryptic as it will require unpickling columns to understand their content.
- It's backwards incompatible. A database must not have any rows when restarting scrapyd to upgrade.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -11,6 +11,16 @@ The highlight of this release is the long-awaited Python 3 support.
 The new scrapy requirement is version 1.0 or higher.
 Python 2.6 is no longer supported by scrapyd.
 
+Some unused sqlite utilities in are now deprecated
+and will be removed from a later scrapyd release.
+Instantiating them or subclassing from them
+will trigger a deprecation warning.
+These are located under ``scrapyd.sqlite``:
+- SqliteDict
+- SqlitePickleDict
+- SqlitePriorityQueue
+- PickleSqlitePriorityQueue
+
 Added
 ~~~~~
 
--- a/scrapyd/_deprecate.py
+++ b/scrapyd/_deprecate.py
@@ -17,7 +17,7 @@ class WarningMeta(type):
         if offending_classes:
             warnings.warn(
                 '%r inherits from %r which %s deprecated'
-                ' and will be removed from a later scrapyd version'
+                ' and will be removed from a later scrapyd release'
                 % (cls, offending_classes,
                    ['is', 'are'][min(2, len(offending_classes))-1]),
                 ScrapydDeprecationWarning,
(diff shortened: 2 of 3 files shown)


Digenis:
Although scrapy has some utils for deprecating a renamed/replaced class,
it doesn't have such for classes that will be removed.
Any advice from the scrapy maintainers?

redapple:
Right, I don't see a utility for to-be-removed classes in scrapy.
w3lib has [explicit `warning.warn` calls](https://github.com/scrapy/w3lib/search?utf8=%E2%9C%93&q=%22will+be+removed+in+future%22&type=Code).

Probably @kmike or @dangra have some good advice to provide here.

Digenis:
w3lib does it only for functions.
For classes, we need warnings not only when initializing
but also when subclassing.

scrapy.utils.deprecate.create_deprecated_class does this
but it's specific to renaming/replacing.
Patching it in scrapy doesn't suffice
because it won't be there for its previous version,
which we are going to support.

I think scrapy.utils.deprecate deserves its own repo.
Pypi lacks a descent package for this kind of functionality.
@dangra has already reserved a name in pypi
but doesn't seem to have uploaded anything.

Although I don't feel like copying it here and modifying it
I don't think yac shaving can yield a realistic release date
so I'll have my take in this PR.

Digenis:
I tried refactoring scrapy.utils.deprecate.create_deprecated_class
to fit it for this purpose but I didn't succeed.
Maybe someone more familiar with metaclasses can help.

Digenis:
I tried writing a metaclass from scratch which worked.
Then I rebased on master where 7d4149e635361549959fb0db06eb36737d6aca90
broke the metaclass thing because of parent classes' metaclasses.
So I wrote a class decorator that defines a new metaclass
based on the class's bases' metaclasses and my metaclass
and uses it as a metaclass for the redefined class.

It still doesn't work when inheriting from multiple deprecated classes.
TravisCI also throws errors that don't occur in my python.

Digenis:
This ticket is stalled because I get a metaclass conflict when inheriting from 2 deprecated classes.
```
Error when calling the metaclass bases
metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```
If I don't manage to solve this
we can just ignore this bug.
There's no reason to inherit from a combination of any of the deprecated classes
because they are incompatible with each other anyway.

Digenis:
@redapple, what do you think of the changelog entry?
I find it too verbose.
We need should we document the deprecation
but should we document the bug? should we justify not fixing it?

Digenis:
For the record,
I tried http://code.activestate.com/recipes/204197-solving-the-metaclass-conflict/
and it does seem to work.
However I feel reluctant to just plug some code without understanding the issue first.

redapple:
My suggestions is to list those "sqlite utilities",
and simply mention that using them will trigger a deprecation warning.

I would not mention the bug for this rare (and probably not realistic) case of inheriting 2 of from 2 of those deprecated classes.
We could amend and document this if users do run into trouble.

Digenis:
I'm going to test everything and wrap up the new release by tomorrow.

redapple:
Awesome @Digenis ! Looking forward to the release.

PR #167, opened 2016-07-28
PY3 Fix crawl arguments parsing

redapple:
I had to copy scrapy's deprecated `stringify_dict` and made it use "native" str (`str` in Python 2, Unicode in Python 3)


Digenis:
Great. This should also fix the deprecation warning.

PR #169, opened 2016-08-01
Drop logpath formatter

Digenis:
This reverts all 3 commits related to #28 #71.
I'm trying to move it to a separate webservice
only not to disappoint scrapyd users
who fell for this feature without waiting for its release.

However, to conclude it,
the config object should be permanently tied to one of the components.
I think people who write custom websites and webservices
should have the ability to define their own config keys.
This would enable them to do so but should be on a separate PR.

**P.S. After trying several things, I gave up moving it to a separate webservice.**

Cc: @m-vdb, @sujaymansingh, @decaz

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -68,23 +68,6 @@ storing logs set this option empty, like this::
 
     logs_dir =
 
-logs_filename
--------------
-
-The filename (appended to ``logs_dir``) to use for the crawl. ``{project}``
-and ``{spider}`` placeholders will be substituted, as will certain datetime
-elements (but only ``{Y}``, ``{m}``, ``{d}``, ``{H}``, ``{M}``, ``{S}``).
-
-For example,
-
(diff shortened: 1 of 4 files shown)


Digenis:
This requires some new features which I'd rather not rush in 1.2.
I suggest we merge only the reverts (f0115c2, db87e84, 7560241).

Anyone who still needs the logs_filename feature can do it on the client side:
```sh
curl http://example.com:6800/schedule.json -d project=project1 -d spider=spider1 -d setting=LOG_FILENAME=$(date +%Y/%m/%d-)$(cat /proc/sys/kernel/random/uuid).log
```
You can also try using @sujaymansingh's fork at https://github.com/sujaymansingh/scrapyd
Since a person who already depends on this feature must have installed scrapyd from github,
it should be easy to just change the git remote in the pip-install command.

Digenis:
Any comments before I merge the reverts?

PR #174, opened 2016-08-30
Don't use Twisted 16.4 (for now) because of removed bin/trial

redapple:
`bin/trial` was removed in Twisted 16.4
https://twistedmatrix.com/trac/ticket/8765
Stick to Twisted 16.3.2 for now


redapple:
Superceded by https://github.com/scrapy/scrapyd/pull/175

PR #175, opened 2016-08-30
[tox] Support Twisted>=16.4 which removed bin/trial

redapple:
Twisted 16.4 removes bin/trial, recommending `python -m twisted.trial`, but I couldn't get it to work for a couple of tests https://travis-ci.org/scrapy/scrapyd/builds/156175309 
I played with some tox config vars but no success (there probably is a way to make it work)

So this PR copies Twisted 16.3.2's bin/trial script into scrapyd to run the tests

### DIFF
--- a/bin/trial
+++ b/bin/trial
@@ -1,0 +1,23 @@
+#!/usr/bin/env python
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
+import os, sys
+
+try:
+    import _preamble
+except ImportError:
+    try:
+        sys.exc_clear()
+    except AttributeError:
+        # exc_clear() (and the requirement for it) has been removed from Py3
+        pass
+
+# begin chdir armor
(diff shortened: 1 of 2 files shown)


Digenis:
It works. Thanks, @redapple

bin/trial
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
```
Digenis: I merged to hastily.
The MIT license requires that it's included with the redistributed software.
Shall we copy its text here as a comment?

redapple:
What about copying Twisted LICENSE in the bin folder?
Also this copy of trial may not qualify as  "substantial portions of the
software" but better be clean

Digenis:
done in 651a2136777e2805e53838f750afdf1106c204c9

PR #181, opened 2016-10-09
Extension of the cancel endpoint.

Rafalonso:
I've been using this myself for a while now but I think the community could appreciate these new cancel endpoints:
- **Cancel**: If we separate cancel into different endpoints it's better to deprecate this one to diminish user confusion.
- **CancelJob**: Exactly the same as Cancel (Ex. `curl http://localhost:6800/cancel.json -d project=myproject -d job=6487ec79947edab326d6db28a2d86511e8247444`)
- **CancelProject**: Cancel every job of a certain project (Ex. curl http://localhost:6800/cancelproject.json -d project=myproject)
- **CancelAll**: Cancel every job. (Ex. `curl http://localhost:6800/cancelall.json`)

Previously I had the cancel.json endpoint decide what to cancel based on the arguments
Ex1 `curl http://localhost:6800/cancel.json -d project=myproject` would cancel all jobs of the project `myproject`
Ex2 `curl http://localhost:6800/cancel.json -d project=myproject -d job=6487ec79947edab326d6db28a2d86511e8247444` would cancel a job
Ex3 `curl http://localhost:6800/cancelall.json` would cancel all jobs

Turned out to be a bad idea for me as I was cancelling too much way too often because I kept forgetting to add an argument. I could also see this being confusing for newcomers. If there's interest in this kind of design I can push it.


Digenis:
This PR introduces too much code repetition.
Can you put it all in a single webservice?

Rafalonso:
I'll give it a go today, will push it as soon as it's done.

Rafalonso:
Ok, I merged the cancel endpoints into 1 webservice. 

To extend the cancel endpoint say, to add cancelspider.json just have to add the function cancelspider into the Cancel webservice and add the endpoint in default_scrapyd.conf

Rafalonso:
Due to inactivity

PR #179, opened 2016-08-30
[WIP] Python 3 port

redapple:
Scrapy 1.1 is required for `to_bytes, to_unicode, to_native_str`

Major changes:
- quite a few `zope.interface.implements` --> `zope.interface.implementer`
- introduce a `native_stringify_dict` to help with interpreting URL parameters among others
- `DictMixin` replaced with `MutableMapping` for sqlite backend
- use paths as bytes when registering children resource for Home resource
- explicit UTF-8 encoding for HTTP responses

PR #185, opened 2016-10-24
Lower scrapy 1.1 dependency to 1.0

Digenis:
An attempt for https://github.com/scrapy/scrapyd/pull/179#issuecomment-243439640

It's hard to verify the return types of all these functions for both pythons
so I did the easy parts in `sqlite.py` and replaced `to_native_str()` in `native_stringify_dict()`

### DIFF
--- a/scrapyd/sqlite.py
+++ b/scrapyd/sqlite.py
@@ -10,9 +10,6 @@ except ImportError:
     from collections import MutableMapping
 import six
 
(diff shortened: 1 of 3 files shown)


Digenis:
Do we have to note anything about scrapy and the license?

redapple:
How do you mean?

Digenis:
The code, though modified, was copied from scrapy.

redapple:
I think it's just fine

PR #186, opened 2016-10-24
[WIP] turn the config into a component of the app

Digenis:
I want to make the config a component of the app
so that other components will get the config options directly from it.
This should make scrapyd able to reload the config (eg with a SIGHUP)
without the need to restart.


jpmckinney:
Closing PR as a 5-year-old work-in-progress. See issue #210.

PR #194, opened 2016-12-21
Use bumpversion when releasing

redapple:
We've been using [bumpversion](https://pypi.python.org/pypi/bumpversion) for many scrapy-related projects.
This, coupled with automatic PyPI deploys from Travis CI, makes releasing a no-brainer (just the releasing part of course)
what do you think @Digenis ?


redapple:
I quite like the `--dry-run` mode:

```
$ bumpversion --dry-run --verbose minor
Reading config file .bumpversion.cfg:
[bumpversion]
current_version = 1.1.0
commit = True
tag = True
tag_name = {new_version}

[bumpversion:file:scrapyd/VERSION]

Parsing version '1.1.0' using regexp '(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)'
Parsed the following values: major=1, minor=1, patch=0
Attempting to increment part 'minor'
Values are now: major=1, minor=2, patch=0
Dry run active, won't touch any files.
Parsing version '1.2.0' using regexp '(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)'
Parsed the following values: major=1, minor=2, patch=0
New version will be '1.2.0'
Asserting files scrapyd/VERSION contain the version string:
Found '1.1.0' in scrapyd/VERSION at line 0: 1.1.0
Would change file scrapyd/VERSION:
--- a/scrapyd/VERSION
+++ b/scrapyd/VERSION
@@ -1 +1 @@
-1.1.0
+1.2.0
Would write to config file .bumpversion.cfg:
[bumpversion]
current_version = 1.2.0
commit = True
tag = True
tag_name = {new_version}

[bumpversion:file:scrapyd/VERSION]


Would prepare Git commit
Would add changes in file 'scrapyd/VERSION' to Git
Would add changes in file '.bumpversion.cfg' to Git
Would commit to Git with message 'Bump version: 1.1.0 → 1.2.0'
Would tag '1.2.0' in Git
```

Digenis:
I don't feel like these are too many commands to write manually
but since this PR is just a config file for bump version, I'm OK with it.

It's usually documentation that gets forgotten.
E.g. the current `doc/config.py` has hardcoded old versions.
The release date should be added in the changelog.
Sometimes we may note the first version where a feature appeared
and sometimes we should remove the note
(e.g. because it's older than the separation of scrapyd from scrapy)

redapple:
True, bumpversion will not update the changelog and this has to be done manually before running bumpversion. We can add `docs/conf.py` so as not to be forgotten. Shall I add that?

redapple:
@Digenis , I added `docs/conf.py` in the list of files, with the simplification that `version` and `release` will be the same

Digenis:
LGTM.
Merge?

redapple:
Done :)

Digenis:
@redapple what do you think of https://github.com/Digenis/scrapyd/commit/0b48e966ae4bd6253d85957a88d98a2e4d4c9636
I want to add alpha/beta/rc suffixes to bumpversion in order to release an alpha.
(To me, bumpversion still doesn't make sense and it would be easier to do it manually)

redapple:
@Digenis , I must say that for Scrapy we were not able to make bumpversion work with alpha/beta/rc (see https://github.com/scrapy/scrapy/pull/2159)
But if https://github.com/Digenis/scrapyd/commit/0b48e966ae4bd6253d85957a88d98a2e4d4c9636 works, I'm fine with it.
Also, if bumpversion does not make sense to you, feel free to remove it. It's no big deal.

Digenis:
I only managed to do it with `bumpversion --new-version 1.2.0alpha1 prerel` (and 1.1.1post1 postrel)
but I think it's meant to be simpler, shouldn't it?

In any case, I can always set the versions manually in all 3 files (.bumpversion.cfg included).

redapple:
For what I use bumpversion for (with the various scrapy-plugins and scrapy/ libraries), with the defaults (no alpha/beta/rc etc), bumpversion is simple enough.
For the "advanced" stuff, I find bumpversion not straightforward. I still like it a lot though.
But it may not be for scrapyd. If you're fine with manual editing of versions and tagging, I don't mind not having bumpversion, really.

PR #195, opened 2016-12-21
[Travis CI] Add auto-deploy to PyPI on master branch tags

redapple:
Reference: https://docs.travis-ci.com/user/deployment/pypi/

Generated with `travis setup pypi`
```
$ travis setup pypi
Detected repository as scrapy/scrapyd, is this correct? |yes| yes
Username: redapple
Password: ****************
release only tagged commits? |yes| 
deploy as wheel file too? |yes| 
Release only from scrapy/scrapyd? |yes| 
Release from travis2pypi-deploy branch? |yes| no
Encrypt Password? |yes| yes
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,5 +53,7 @@ deploy:
     password:
         secure: IuHONsIRm8w54LkmUlZe0X0dFqW2iAjTyqeRyOyVMvHWIPaiydNpX+xipvRArJnswlnzf7/X3CtZw9ZF9ePznCrvG+3Y1SLHutncRbr6IxSQrcRJm+FTF2I2KNEaOkWqoQjO7xbEehpqo6fttHImCNrCzgZw32hCZqeGbDGPGJs=
     on:
+        all_branches: true
         tags: true
         repo: scrapy/scrapyd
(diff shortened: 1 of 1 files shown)


Digenis:
Are you sure about the "on master branch tags" part?
Because it can only mean "tags referenced by master"
which means that patched previous releases won't be uploaded to pypi

redapple:
This is intended for master branch. but feature branch can have this with some adaptation.
Scrapy's config has a regex to validate the tag: https://github.com/scrapy/scrapy/blob/master/.travis.yml#L38

Digenis:
The current git workflow
has a master branch for development
PRs for all kinds of updates
and release branches for bugfixes/backports.
To my understanding, it's the same for scrapy.
If I got this correctly,
to upload an old release branch which just go a minor bump
we need: `condition: "$TRAVIS_BRANCH =~ ^[0-9]+[.][0-9]+$"`

redapple:
Truth is we started to release from master branch also. (One of the reasons is that we had branched out very early for 1.1, and it was a real pain to backport everything when master and 1.1 has no difference.)
But I get your point.
I quite like the tag test of scrapy's config vs. the branch test you propose. But you decide.

Digenis:
The tag test will also work if the commit is not referenced by master, right?
If so, let's add the tag test too and merge.

redapple:
@Digenis , I added the "all_branches: true" bit to be sure it doesn't restrict to "master" + the tag test.

Digenis:
LGTM.
Merge?

Issue #40, opened 2014-03-08
SqlitePriorityQueue.pop() return None may crash Poller.poll() 


aspidites:
Continued discussion of scrapy/scrapy#28 here since it is directly related to scrapyd.

I can think of a few different ways to handle this, but until I get more familiar with scrapy(d)'s internals, I'm not sure which is preferable:
1. Impement for fine-grained control over sqlite3's transactions
2. Increase the timeout limit
3. Test for an ongoing transaction before hand

I suspect 1 would be most robust but 3 would be least intrusive.

Is it enough to change `scrapyd/sqlite.py` line 14 to:

``` pyhon
self.conn = sqlite3.connect(self.database, isolation_level=None, check_same_thread=False)
```

? If so, we could also remove explicit calls to `commit()` since they would be committed automatically after `execute()`

Digenis:
I quote here the original ticket to make it easier to read all the details in one place

@shaneaevans wrote:
>Previously reported by mrkschan on Trac http://dev.scrapy.org/ticket/313

>I have a scrapy project that has several spiders in it. Those spiders are scheduled to execute on an hourly-basis.

>The scrapyd eventually report unhandled error (as shown below). When I dig through the source of scrapy, I suspect the concurrent control of sqlite3 access is not well guarded as the error below is caused by popping from an empty queue.

>The case can simply be explained by the scenario with two spiders A and B (also refer to the source - http://is.gd/ht4HSs).

>1. Spider A's poller poll() get to line 21 of scrapyd/poller.py.
2. Meanwhile, Spider B's poller also get to line 21.
3. Spider A's poller poll() get to line 23, the queue in sqlite becomes empty and sqlite3 is locked (according to py doc - http://is.gd/67PKpn).
4. Spider B's poller poll() also get to line 23, wait the lock to be released.
5. Spider A commit and release the sqlite lock.
6. Spider B pop() from an empty queue. Raise Error.

>```
Traceback (most recent call last):
  File "/usr/lib/python2.6/dist-packages/twisted/internet/base.py", line 778, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/usr/lib/python2.6/dist-packages/twisted/internet/task.py", line 194, in __call__
    d = defer.maybeDeferred(self.f, *self.a, **self.kw)
  File "/usr/lib/python2.6/dist-packages/twisted/internet/defer.py", line 117, in maybeDeferred
    result = f(*args, **kw)
[...]

PR #198, opened 2017-01-03
fix poller race condition

Digenis:
This is an easy fix for #40.

The queue and the poller do need some refactoring
but we'll do this on some later version.


Digenis:
I tried writing a test for it but it needs deep mocking.

In the default scrapyd setup
queue.pop theoretically will never be called on an empty queue
so we already have enough coverage.

@bezkos already tested it https://github.com/scrapy/scrapyd/issues/40#issuecomment-270107672

@redapple, comments?

redapple:
LGTM

Issue #197, opened 2016-12-26
Configurable spider queue class


bezkos:
I have started to implement a custom job queue so i can use a shared job queue with Postgre.
I tried to use this setting SPIDER_QUEUE_CLASS but i found out that its missing from Scrapyd. Is it possible to have this setting back cause i want to avoid patch scrapyd code? I think implementation of this feature again, it is important when someone want to use Scrapyd in a multi-server environment.

Digenis:
I summarize here my reply from the mailing list (only for the record).

The `SPIDER_QUEUE_CLASS` setting
is from the time that scrapyd was part of scrapy
not only as a package but as a module.
https://github.com/scrapy/scrapy/commit/75e2c3eb338ea03e487907fa8c99bb12317e9435
Unfortunately the release notes do not cover all the details of scrapyd's separation as a module
(and it would probably be impractical)

Scrapyd never had this setting.

Digenis:
It seems easy to implement such a setting.
```patch
diff --git a/scrapyd/default_scrapyd.conf b/scrapyd/default_scrapyd.conf
index 0da344f..e2b0c35 100644
--- a/scrapyd/default_scrapyd.conf
+++ b/scrapyd/default_scrapyd.conf
@@ -15,4 +15,5 @@ runner      = scrapyd.runner
 application = scrapyd.app.application
 launcher    = scrapyd.launcher.Launcher
+spiderqueue = scrapyd.spiderqueue.SqliteSpiderQueue
 webroot     = scrapyd.website.Root
 
diff --git a/scrapyd/utils.py b/scrapyd/utils.py
index 602a726..c96add0 100644
--- a/scrapyd/utils.py
+++ b/scrapyd/utils.py
@@ -9,8 +9,11 @@ import json
 from twisted.web import resource
 
-from scrapyd.spiderqueue import SqliteSpiderQueue
+from scrapy.utils.misc import load_object
 from scrapyd.config import Config
 
 
+DEFAULT_SPIDERQUEUE = 'scrapyd.spiderqueue.SqliteSpiderQueue'
+
+
 class JsonResource(resource.Resource):
 
@@ -57,8 +60,9 @@ def get_spider_queues(config):
     if not os.path.exists(dbsdir):
         os.makedirs(dbsdir)
+    spiderqueue = load_object(config.get('spiderqueue', DEFAULT_SPIDERQUEUE))
     d = {}
     for project in get_project_list(config):
[...]

Issue #187, opened 2016-11-02
Polling order


Digenis:
The poller demonstrates "preferences" for some projects over others
and lets them monopolize process slots.

The following code in `poller.py`
polls projects in whatever order `iteritems()` uses
which is arbitrary but constant.
```python
for p, q in self.queues.iteritems():
    c = yield maybeDeferred(q.count)
    if c:
        msg = yield maybeDeferred(q.pop)
        returnValue(self.dq.put(self._message(msg, p)))
```
This becomes obvious when multiple projects
have several spiders queued and waiting for slots.
The first `(project, queue)` pair returned from `iteritems()`
that has queued spiders
will prevent any further projects from being polled.
If its queue is constantly full
only this project will have its spiders scheduled.

Possible solutions:

One way to solve this is suggested by @luc4smoreira in #140.
However it can't be a correct solution (for this issue)
because it can make total slot utilization vary a lot.
E.g. with 5 slots per project and 10 projects,
when 1 project is active there will be only 5 slots utilized
(even if the single active project is in need for more)
but when all 10 are active, there will be 50 slots.
In a workaround for the above, one can set the total slot limit
just above the slots per project
multiplied by the number of projects expected to monopolize slots
so that there will always be a few slots for others.

Another incomplete solution is to use the spider priority earlier in the poller
before picking a project. When I noticed this problem, I first tried lowering the priority
of spiders belonging to the project monopolizing the slots
and was surprised because it had no effect.

We can also stop using multiple queues
and keep everything in a single database.
Although it sounds like a big change,
by doing it we'll drop a lot of code from scrapyd
and we can even implement both of the 2 suggestions above
cleaner, faster and with less code.

Digenis:
The easiest way to fix this
is pick the queue to poll at random
instead of the iteritems order,
like in the following dirty workaround:

```patch
--- a/scrapyd/poller.py
+++ b/scrapyd/poller.py
@@ -1,3 +1,4 @@
+from random import sample
 from zope.interface import implements
 from twisted.internet.defer import DeferredQueue, inlineCallbacks, maybeDeferred, returnValue
 
@@ -17,11 +18,12 @@ class QueuePoller(object):
     def poll(self):
         if self.dq.pending:
             return
-        for p, q in self.queues.iteritems():
-            c = yield maybeDeferred(q.count)
-            if c:
-                msg = yield maybeDeferred(q.pop)
-                returnValue(self.dq.put(self._message(msg, p)))
+
[...]

PR #201, opened 2017-01-06
configurable spider queue class

Digenis:
(TODO: doc)
This was once (scrapy<0.14) supported by a scrapy setting.
I have plans for merging the sqlite priority queue into the spider queue
but this change is unlikely to interfere.

Full context: #197 
Cc: @bezkos

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -137,6 +137,19 @@ Scrapyd includes an interface with a website to provide simple monitoring
 and access to the application's webresources.
 This setting must provide the root class of the twisted web resource.
 
+.. _spiderqueue:
+
+spiderqueue
+-----------
+
+The per-project spider queues are where the scheduler enqueues crawls
+for the poller to pick.
+You can define a custom spider queue class here
+as long as it implements the ISpiderQueue interface.
+
+Defaults to ``scrapyd.spiderqueue.SqliteSpiderQueue``.
+
+
 node_name
 ---------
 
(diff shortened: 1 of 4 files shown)


Digenis:
I just completed this PR
but now I have second thoughts.

If we expose the spider queue,
then one discussed solution for #187,
(unifying queues/dbs for deployed projects)
will need deprecation cycles.

I'm moving this to the 1.3 milestone.

@bezkos, I'm interested in hearing your opinion on this.

Digenis:
postponing this in favour of #187/3rd solution (Unify queues/dbs)

mxdev88:
> postponing this in favour of #187/3rd solution (Unify queues/dbs)

hi @Digenis, it shouldn't be an issue to make queues configurable even if one wants to use other projects mentioned in #187. This PR would actually be an enabler. Any chance to see this merged anytime soon?

Digenis:
Hi @mxdev88,

I haven't been using scrapyd for some years now
and continuing development without actually using it
feels like taking decisions for others with nothing at stake.
This project needs maintainers.

Regarding this specific PR,
I think the problem was that after #187 
this feature's code would need a revisit
and anyone using it would also need to reconsider changes from #187.
The problem with #187 is that there are multiple queuing strategies,
each one valid for different use cases.
Maybe the poller and the queue should be merged in a single class.

jpmckinney:
New PR without conflicts: #476

PR #202, opened 2017-01-13
Add start/finish time columns to job list

vimagick:
Displaying start/finish time on job list is very handy for schedule checking.

screen shot 2017-01-14 at 01 06 47

### DIFF
--- a/scrapyd/website.py
+++ b/scrapyd/website.py
@@ -111,7 +111,7 @@ class Jobs(resource.Resource):
         s += "<h1>Jobs</h1>"
         s += "<p><a href='..'>Go back</a></p>"
         s += "<table border='1'>"
-        s += "<tr><th>Project</th><th>Spider</th><th>Job</th><th>PID</th><th>Start</th><th>Finish</th><th>Runtime</th><th>Log</th>"
+        s += "<tr><th>Project</th><th>Spider</th><th>Job</th><th>PID</th><th>Start</th><th>Runtime</th><th>Finish</th><th>Log</th>"
         if self.local_items:
             s += "<th>Items</th>"
             cols = 9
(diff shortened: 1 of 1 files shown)


scrapyd/website.py
```diff
@@ -140,7 +142,9 @@ def render(self, txrequest):
             for a in ['project', 'spider', 'job']:
                 s += "<td>%s</td>" % getattr(p, a)
             s += "<td></td>"
-            s += "<td>%s</td>" % (p.end_time - p.start_time)
+            s += "<td>%s</td>" % p.start_time.replace(microsecond=0)
+            s += "<td>%s</td>" % p.end_time.replace(microsecond=0)
+            s += "<td>%s</td>" % (p.end_time.replace(microsecond=0) - p.start_time.replace(microsecond=0))
```
Digenis: `(p.end_time - p.start_time).replace(microsecond=0)`
floor/round/ceil for display purposes should only happen at the end of the calculation

scrapyd/website.py
```diff
@@ -129,7 +129,9 @@ def render(self, txrequest):
             s += "<tr>"
             for a in ['project', 'spider', 'job', 'pid']:
                 s += "<td>%s</td>" % getattr(p, a)
-            s += "<td>%s</td>" % (datetime.now() - p.start_time)
+            s += "<td>%s</td>" % p.start_time.replace(microsecond=0)
+            s += "<td></td>"
+            s += "<td>%s</td>" % (datetime.now().replace(microsecond=0) - p.start_time.replace(microsecond=0))
```
Digenis: https://github.com/scrapy/scrapyd/pull/202/commits/5893965106b1ee6947ed42be01e647fbcf1d91dc#r96112811 applies here too

Digenis:
Hi @vimagick,
thank you for your contribution.

I think "start, runtime, finish" is a more intuitive column order.
I'm not sure I mind the missing microseconds
as long as they are available in the json api (the webservices).
I do use a similar patch and find it very useful.

We should remember to consider adding the enqueue time,
if we start tracking it too. Currently we don't.

What do other project member think?
Should we include it in 1.2?

scrapyd/website.py
vimagick: @Digenis No such method: `datetime.timedelta.replace`

Digenis: Well, python just surprised me in a bad way :confused:
We could solve this with a rounding/truncating function
but let's not bother. The current precision suffices.

Digenis:
Will merge if no comments.
@vimagick, how about putting the runtime column before the finish column?

vimagick:
@Digenis No problem 👌

PR #203, opened 2017-01-19
native string type in exc msg in get_spider_list

Digenis:
Exception messages are unicode in py3 and bytes in py2.
If unicode characters are needed in py2, they should be encoded to unicode_escape.

We did this before in scrapy: https://github.com/scrapy/scrapy/pull/952

This is a trivial fix, we can do a fast-forward merge


redapple:
Looks good to me

Issue #70, opened 2014-12-11
Use --rundir when parsing relative *_dir options to absolute paths


sujaymansingh:
Hi

I run scrapyd with the `--rundir` option. (version 1.0.1)

I have the following issue.
- I deploy an egg to scrapyd
- egg and project are present (I can see them in `listprojects.json`)
- I then restart scrapyd
- egg and project are not present in `listprojects.json`
- BUT: egg file is present in `$rundir/eggs`

I suspect the issue is to do with **when** the directory is changed.

It seems that `SpiderScheduler` will load the eggs/projects when initialised
https://github.com/scrapy/scrapyd/blob/1.0.1/scrapyd/scheduler.py#L12

But I think this is done **before** changing the working directory to `--rundir`.

To investigate, I added a couple of hacky print statements to `SpiderScheduler`

``` python
class SpiderScheduler(object):

    implements(ISpiderScheduler)

    def __init__(self, config):
        self.config = config
        import os; print "SpiderScheduler::__init__ current dir=" + os.getcwd()
        self.update_projects()

    def schedule(self, project, spider_name, **spider_args):
        q = self.queues[project]
        q.add(spider_name, **spider_args)

    def list_projects(self):
        import os; print "SpiderScheduler::list_projects current dir=" + os.getcwd()
        return self.queues.keys()

    def update_projects(self):
        self.queues = get_spider_queues(self.config)
```

And grepping logs for "SpiderScheduler" (after I restart and then make a call in my browser to `listprojects.json`)

```
SpiderScheduler::__init__ current dir=/opt/skuscraper
2014-12-11 15:22:40+0000 [HTTPChannel,0,10.10.9.220] SpiderScheduler::list_projects current dir=/var/scrapyd
```

So `/opt/skuscraper` is my project directory (with the `scrapyd.conf`).
But I want the working directory to be separate (so it doesn't put any extra files in the app directory), that is why I use `/var/scrapyd` as the run dir.

We can see that when the SpiderScheduler object is init'd, the current dir is `/opt/skuscraper`, so it can't find any eggs.
But after the app starts up, it uses `/var/scrapyd`.

So any deploys of eggs **after** the app starts up are saved to `/var/scrapd/eggs`, but then scrapyd is restarted, it loads its initial list of eggs from `/opt/skuscraper` (where they won't exist).

Digenis:
See https://twistedmatrix.com/trac/ticket/2572
Does updating twisted solves this?
[...]

PR #209, opened 2017-02-03
Update config.rst

funkyfuture:
i'm emphasizing two aspects here that took me a while to figure out.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -10,11 +10,15 @@ them in order with the latest one taking more priority:
 * ``c:\scrapyd\scrapyd.conf`` (Windows)
 * ``/etc/scrapyd/conf.d/*`` (in alphabetical order, Unix)
 * ``scrapyd.conf``
-* ``~/.scrapyd.conf`` (users home directory)
+* ``~/.scrapyd.conf`` (user's home directory)
 
 The configuration file supports the following options (see default values in
 the :ref:`example <config-example>`).
 
+.. note::
+    The user's home directory is not expanded for the user that may be specified
+    with the ``--uid`` option, but for the user invoking ``scrapyd``.
+    
 http_port
 ---------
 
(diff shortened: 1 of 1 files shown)


docs/config.rst
```diff
@@ -10,11 +10,15 @@ them in order with the latest one taking more priority:
...
-* ``~/.scrapyd.conf`` (users home directory)
+* ``~/.scrapyd.conf`` (user's home directory)
 
 The configuration file supports the following options (see default values in
 the :ref:`example <config-example>`).
 
+.. note::
+    The user's home directory is not expanded for the user that may be specified
+    with the ``--uid`` option, but for the user invoking ``scrapyd``.
+
```
Digenis: The --uid option is for twistd, not scrapyd.
I'm not sure that you are supposed to pass options to twistd.
There may be problems such as #70

Your note makes sense
but it may give the impression
that passing additional arguments to twistd
is going to work as expected.
Have you tried using other twistd arguments?

Digenis:
Hi @funkyfuture, welcome to scrapyd.
Thank you for tiding up the documentation.
We will merge this but I want to check some things first (see my comment on the code)

docs/config.rst
funkyfuture: yes, i also use `--nodaemon`. `--debug` also seems to work, but i haven't tested to drop into the pdb.

if passing arguments to twisted wasn't supposed, that should be actual behaviour, including not to show twisted's `--help`.

Digenis: TLDR: Why do you need `--uid`? Can we find at which point twistd sets it? 

It looks like you are using scrapyd with `twistd -y`
instead of the provided console script (which uses `--nodaemon` by default).
If we do our own argument parsing in the console script (eg to solve #70)
the console script may start to become the only supported way to run scrapyd.
There's also `scrapyd.script` which I think we should deprecate.

Why do you need to set the uid in the first place?
In case you start as root and drop privileges,
why would you need root?

I can't yet tell at which point twistd drops privileges (provided a uid option)
but if things are as you describe them in the patch, it's definitely after reading the config.

In #186, I want to enable scrapyd to reload the config
without restarting (and killing all running spiders).
I want to first try reading the config after changing the uid
(i.e. render your note false) so that the above feature
will be able to be combined with the uid option.

funkyfuture: in short: i don't need to invoke scrapyd as root, but i usually check out a command's help and spotted that option and chose to go with it. then it took me a while to figure out why `scrapyd` wouldn't read the config from the `--uid` defined user's directory.

i invoke `scrapyd` just as that, which points to [`scrapyd.scripts.scrapyd_run:main`](https://github.com/scrapy/scrapyd/blob/master/setup.py#L47). if one was supposed to use another script, then that entrypoint shouldn't be defined at all.

shall i open a report that the `--help` command doesn't what it should do?

Issue #231, opened 2017-05-03
error when deploying  !!!


crisfan:
when I was deploying my project, there was a error that I coudn't resovle!
error message:
{"status": "error", "message": "environment can only contain strings", "node_name": "XSOOY-PC"}

server message:
```
 Traceback (most recent call last):
          File "c:\program files (x86)\python27\lib\site-packages\twisted\web\http.py", line 1694, in allContentReceived
            req.requestReceived(command, path, version)
          File "c:\program files (x86)\python27\lib\site-packages\twisted\web\http.py", line 790, in requestReceived
            self.process()
          File "c:\program files (x86)\python27\lib\site-packages\twisted\web\server.py", line 189, in process
            self.render(resrc)
          File "c:\program files (x86)\python27\lib\site-packages\twisted\web\server.py", line 238, in render
            body = resrc.render(self)
        --- <exception caught here> ---
          File "c:\program files (x86)\python27\lib\site-packages\scrapyd\webservice.py", line 21, in render
            return JsonResource.render(self, txrequest).encode('utf-8')
          File "c:\program files (x86)\python27\lib\site-packages\scrapyd\utils.py", line 20, in render
            r = resource.Resource.render(self, txrequest)
          File "c:\program files (x86)\python27\lib\site-packages\twisted\web\resource.py", line 250, in render
            return m(request)
          File "c:\program files (x86)\python27\lib\site-packages\scrapyd\webservice.py", line 86, in render_POST
            spiders = get_spider_list(project, version=version)
          File "c:\program files (x86)\python27\lib\site-packages\scrapyd\utils.py", line 132, in get_spider_list
            proc = Popen(pargs, stdout=PIPE, stderr=PIPE, env=env)
          File "c:\program files (x86)\python27\lib\subprocess.py", line 390, in __init__
            errread, errwrite)
          File "c:\program files (x86)\python27\lib\subprocess.py", line 640, in _execute_child
            startupinfo)
        exceptions.TypeError: environment can only contain strings
```

what should I do?

Digenis:
@crisfan it could be similar to #212.
Do you have a version scheme configured in your scrapy.cfg?

crisfan:
@Digenis， I think my problem  not be similar to #212 .
this is my scrapy.cfg
[settings]
default = zzh.settings

[deploy:127]
url = http://localhost:6800/
project = zzh

when I deploy my project, I receive this version:
Packing version 1493965313
[...]

PR #242, opened 2017-09-01
Native-stringify project and version arguments when deploying

redapple:
Should fix one case of https://github.com/scrapy/scrapyd/issues/231.
This does not fix the case where [`SCRAPY_PROJECT` is set to a `dict`](https://github.com/scrapy/scrapyd/issues/231#issuecomment-315743172).

I was able to reproduce `exceptions.TypeError: environment can only contain strings` with Miniconda and Python 2.7 on Windows 7. And the problem was Unicode strings for `project` and `version`.

This patch is similar to what @evrn1874 suggests in https://github.com/scrapy/scrapyd/issues/231#issuecomment-326577064 , but it tries to use the same pattern as other `render_POST()`.

### DIFF
--- a/scrapyd/webservice.py
+++ b/scrapyd/webservice.py
@@ -79,9 +79,10 @@ class Cancel(WsResource):
 class AddVersion(WsResource):
 
     def render_POST(self, txrequest):
-        project = txrequest.args[b'project'][0].decode('utf-8')
-        version = txrequest.args[b'version'][0].decode('utf-8')
-        eggf = BytesIO(txrequest.args[b'egg'][0])
+        eggf = BytesIO(txrequest.args.pop(b'egg')[0])
+        args = native_stringify_dict(copy(txrequest.args), keys_only=False)
+        project = args['project'][0]
+        version = args['version'][0]
         self.root.eggstorage.put(eggf, project, version)
         spiders = get_spider_list(project, version=version)
         self.root.update_projects()


ghostdog2046:
@redapple,Any update for this problem?

redapple:
@ghostdog2046 , no update. This PR still needs review

Digenis:
looks good to me.

I think that if we use native_stringify_dict too much
we may end up hiding other bugs because of
https://github.com/scrapy/scrapyd/blob/891b50c50377f1049a8eb4c4ec8ce70e3eacf987/scrapyd/utils.py#L93-L94
In this case, I don't see how the request would contain anything not a list, dict or string,
but then again, we also don't know how this dict got there.

We could let this PR open until we figure out the other bug
but this fixes a problem for @evrn1874
so let's not keep him waiting.

Sudeepa14:
This fix looks good for #231 and it is still not fixed for windows! any plans on merging?

Digenis:
@my8100, does this work for you?

my8100:
@Digenis I only know that [this pull request](https://github.com/my8100/scrapyd/commit/781bf1ade4e7025c66af664dba27c55e28bc17fa) works for me. 
Why not test other solutions by yourself?

Digenis:
@my8100  https://github.com/my8100/scrapyd/commit/781bf1ade4e7025c66af664dba27c55e28bc17fa
can hide and complicate the dict part of #231

I'm merging this instead although I still have my worries.
In 1.4 we'll do away with all the py2 compatibility.

PR #250, opened 2017-10-13
Drop deprecated classes

Digenis:
This drops the unused sqlite helpers and their tests.

### DIFF
--- a/scrapyd/sqlite.py
+++ b/scrapyd/sqlite.py
@@ -1,8 +1,4 @@
 import sqlite3
-try:
-    import cPickle as pickle
-except:
-    import pickle
 import json
 try:
     from collections.abc import MutableMapping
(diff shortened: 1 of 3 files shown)


Digenis:
There's only one failure related to the build environment configuration.
We use a py3.3 env which doesn't include enum34 (introduced in py3.4)
but cryptography assumes it's there for any py3.
https://github.com/pyca/cryptography/blob/master/setup.py#L283
I'll double check and open an issue there if I'm correct.

Digenis:
https://github.com/pyca/cryptography/blob/a17bcba39a5f0bb0c84a3a4835ed82e023b4dc3f/setup.py#L264
The cryptography module doesn't support python3.3 but scrapy does.
Perhaps we should drop python3.3 support or have a conditional dependency.

This issue is ready for review

Issue #252, opened 2017-10-19
Officially deprecate Ubuntu packages from archive.scrapy.org


boogheta:
Hello,
It seems like the packages source server is not up anymore, resulting in apt updates errors.
Is this a temporary issue or has the server been completely given up? If the latter, it would be good to update scrapyd's public doc as it recommends to install using this repository source.

amin24e:
i'm facing the same problem

Digenis:
Cc: @redapple

redapple:
You can consider archive.scrapy.org as deprecated. I don't forsee it getting up again in the future.

boogheta:
Can you update the public documentation then?
It leads to "apt-get update" errors for users who did use it and it is really misleading letting people think there exists actual package to install the daemon when there aren't anymore :/

redapple:
Right @boogheta . I've renamed the issue here.

Digenis:
From what I read, there's no deprecation.
The mirror is down permanently
so we'd rather remove it from the doc.

Jazzity:
So um, as someone new to all this, what does this mean? Should we just ignore the paragraph "Installing Scrapyd in Ubuntu" and just go for "pip install scrapyd"?

Digenis:
@Jazzity, yes, the paragraph about installing scrapyd in ubuntu should be ignored.
There are no longer instructions for a system install of scrapyd.
However, reading the install scripts for ubuntu can help you figure out how to do a system install.
We need documentation on this. PRs are welcome.

Digenis:
resolved in 60039c76270c7020062bd41919fe0631846db1ca

the should be a separate ticket for adding install documentation

PR #253, opened 2017-10-19
warning for apt-get install scrapyd on ubuntu

mauhai:
As determined over at Scrapy, the apt-get method is based on an outdated package and an install won't work (even if instructions are correctly followed) on Ubuntu 16.04 and above. See https://github.com/scrapy/scrapy/issues/2137 for further information.

I would even suggest to deprecate this part, until the underlying packages have been updated (as done over at the scrapy repo).

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -38,6 +38,10 @@ send a pull request on Github.
 Installing Scrapyd in Ubuntu
 ----------------------------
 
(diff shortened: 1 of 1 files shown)


Digenis:
Actually the ubuntu repo doesn't exist anymore.
I deleted the whole section.
See also #252

Issue #205, opened 2017-01-23
Add schedule.json parameters to listjobs.json response


sieira:
Is there a way to retrieve the parameters of a scheduled task?

Currently
```sh
$ curl http://localhost:6800/schedule.json -d project=myproject -d spider=somespider -d setting=DOWNLOAD_DELAY=2 -d arg1=val1
```
Returns

```json
{"status": "ok",
 "pending": [{"id": "78391cc0fcaf11e1b0090800272a6d06", "spider": "somespider"}]
}
```

I wonder if there is a way to get:

```json
{"status": "ok",
 "pending": [{
    "id": "78391cc0fcaf11e1b0090800272a6d06",
    "spider": "somespider",
    "setting": {"DOWNLOAD_DELAY": 2},
    "arg1": "val1"
}]
}
```

Digenis:
@sieira, no.
This information exists only for pending jobs.
I'd say scrapyd is far from providing this feature.
Also, I think we shouldn't provide an API for this
until we rewrite the webservice to accept parameters in a better format.

my8100:
Try this https://github.com/my8100/scrapydweb , and you can get the schedule history from http://127.0.0.1:5000/schedule/history.log
 
![image](https://user-images.githubusercontent.com/37427675/47090132-d6585880-d254-11e8-80e0-e657f73c8b1a.png)

jabadia:
Would you be willing to accept a patch with this enhancement?
* store the arguments in `ScrapyProcessProtocol()` (around https://github.com/scrapy/scrapyd/blob/4520c45acc115d0fc766eae30788468a77690433/scrapyd/launcher.py#L44)
* report the arguments in `listjobs.json` for pending, running and finished jobs

jpmckinney:
Maybe related: #39

jpmckinney:
> store the arguments in ScrapyProcessProtocol()

Noting that this was the approach in #254

jpmckinney:
See also comments in PR #272

jpmckinney:
Done. Args are collected into `args` instead of being free-floating.

PR #254, opened 2017-10-20
Show all parameters in UI which were passed via -a flag and also have an additional column to "STOP" a particular job

iamumairayub:
I have modified scrapyD UI to show all parameters which were passed to it, and also have an additional column to "STOP" a particular job ...

### DIFF
--- a/scrapyd/website.py
+++ b/scrapyd/website.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 
-import socket
+import socket, json
 
 from twisted.web import resource, static
 from twisted.application.service import IServiceCollection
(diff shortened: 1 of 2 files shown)


scrapyd/website.py
```diff
@@ -105,43 +105,124 @@ def __init__(self, root, local_items):
         self.local_items = local_items
 
     def render(self, txrequest):
-        cols = 8
-        s = "<html><head><title>Scrapyd</title></head>"
+        cols = 0
+
+        s = "<html><head><title>Scrapyd</title>"
+        s += """
+            <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
```
Digenis: The scrapyd website is meant to be minimal.
Using javascript libraries if far from this.
This feature can be implemented with plain forms.
We can even change the webservice to allow get requests
and make it as simple as a link next to each job.

Digenis:
Hi @mani619cash,
this is a useful feature that can be included in 1.3.
However I don't agree with the way it's implemented.

Loading jquery from google is too much of a bloat
compared to the current state of scrapyd
which never used javascript.

See my comment in the code,
it can be implemented in a much simpler way.
If you are not sure, I'll prepare a PR.

jpmckinney:
Closing as abandoned by author. Related: #205

PR #256, opened 2017-11-15
Extend jobtable

Digenis:
This improves the job listing table.
It adds a cancel endpoint (part of #254)
and rewrites the resource class (refactoring).

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -7,6 +7,12 @@ Release notes
 -----
 *Unreleased*
 
+
+Added
+~~~~~
+
+- Jobs website shortcut to cancel a job using the cancel.json webservice.
+
 Removed
 ~~~~~~~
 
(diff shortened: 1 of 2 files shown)


my8100:
@Digenis I install the latest version of Scrapyd via git clone and find that **the cancel button is hidden since its style is set to "display: none;"**. The result is the same when items_dir is not set in the scrapyd.conf file.

image

Digenis:
@my8100, any chance that you deactivated cancel.json
in any of the [scrapyd configuration files](https://github.com/scrapy/scrapyd/blob/19f63b73a5497e8b0c57bd2219894630c14b29e5/scrapyd/config.py#L29-L32)?

Note to self: I think we should list the available webservices
on the webpage's root to ease troubleshooting in issues like this one.

my8100:
#312

PR #260, opened 2017-11-22
Add JobStatus REST web service to easily get the status of a JobEspec…

### LINKED ISSUE #5: Add status.json webservice to view status of specific job

aleroot:
Add a status.json REST endpoint to easily check the status of a single Job.

This feature can be used to poll the web service to notice when a single job has finished.

closes #5


scrapyd/VERSION
```diff
@@ -1 +1 @@
-1.2.0
+1.2.1
```
Digenis: This is an added feature.
We only add features in new releases
which means it will wait until 1.3.
Also, we don't bump versions in master
until it's time for a new release.
This hunk should not be committed.

scrapyd/default_scrapyd.conf
```diff
@@ -19,6 +19,7 @@ webroot     = scrapyd.website.Root
 [services]
 schedule.json     = scrapyd.webservice.Schedule
 cancel.json       = scrapyd.webservice.Cancel
+status.json       = scrapyd.webservice.JobStatus
```
Digenis: better name the endpoint jobstatus to be more explicit
and consistent with other ws names

scrapyd/webservice.py
```diff
@@ -35,7 +35,6 @@ def render_GET(self, txrequest):
 
         return {"node_name": self.root.nodename, "status":"ok", "pending": pending, "running": running, "finished": finished}
 
-
```
Digenis: This makes the style more consistent with the blank lines in the rest of the file
but is a counter-pep8 whitespace change that shouldn't be commited.
Perhaps I should fix all blank lines in a single commit.

Digenis:
Hi @aleroot,

welcome to scrapyd

I added commented on some lines.
Commit messages should be confined to 50 characters or less.
I still haven't reviewed the whole patch, I'll finish it tomorrow.

aindong:
Any news on this feature? Can we filter by jobid instead of returning the whole list of jobs?

psdon:
any news with this PR?

Digenis:
I think only docs/api.rst needs to be updated

it's a small change so it's possible to include it in 1.3

scrapyd/default_scrapyd.conf
jpmckinney: We don't call the others `jobschedule` or `jobcancel` so I think I'll return this to `status`.

Issue #246, opened 2017-09-27
scrapyd cant install install_requires from setup.py


junejie:
I build the egg using `scrapyd-deploy --build-egg=dist/articles.egg`

I deploy it using curl and it trow error.

Error: 

```
Traceback (most recent call last):
      File "/User/accountname/lib/python3.5/site-packages/twisted/web/http.py", line 2104, in allContentReceived
        req.requestReceived(command, path, version)
      File "/User/accountname/lib/python3.5/site-packages/twisted/web/http.py", line 866, in requestReceived
        self.process()
      File "/User/accountname/lib/python3.5/site-packages/twisted/web/server.py", line 195, in process
        self.render(resrc)
      File "/User/accountname/lib/python3.5/site-packages/twisted/web/server.py", line 255, in render
        body = resrc.render(self)
    --- <exception caught here> ---
      File "/User/accountname/lib/python3.5/site-packages/scrapyd/webservice.py", line 21, in render
        return JsonResource.render(self, txrequest).encode('utf-8')
      File "/User/accountname/lib/python3.5/site-packages/scrapyd/utils.py", line 20, in render
        r = resource.Resource.render(self, txrequest)
      File "/User/accountname/lib/python3.5/site-packages/twisted/web/resource.py", line 250, in render
        return m(request)
      File "/User/accountname/lib/python3.5/site-packages/scrapyd/webservice.py", line 86, in render_POST
        spiders = get_spider_list(project, version=version)
      File "/User/accountname/lib/python3.5/site-packages/scrapyd/utils.py", line 137, in get_spider_list
        raise RuntimeError(msg.encode('unicode_escape') if six.PY2 else msg)
    builtins.RuntimeError: Traceback (most recent call last):
      File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/runpy.py", line 193, in _run_module_as_main
        "__main__", mod_spec)
      File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/runpy.py", line 85, in _run_code
        exec(code, run_globals)
      File "/User/accountname/lib/python3.5/site-packages/scrapyd/runner.py", line 40, in <module>
        main()
      File "/User/accountname/lib/python3.5/site-packages/scrapyd/runner.py", line 37, in main
        execute()
      File "/User/accountname/lib/python3.5/site-packages/scrapy/cmdline.py", line 109, in execute
        settings = get_project_settings()
[...]

Issue #224, opened 2017-04-06
Run crawls within virtualenvs


redapple:
Different projects may have different dependencies, and commonly enough, different Python dependencies.

What about:
- specifying a list of Python dependencies at deploy time, for example in the [well-known requirements file](https://pip.pypa.io/en/stable/user_guide/#requirements-files) format
- having scrapyd create a virtualenv, and installing with these dependencies, either at deploy time, or at crawl run time,
- run crawls withing this virtualenv

PR #269, opened 2018-01-16
auto pip install requires.(install_requires of setup.py)

cleocn:
https://github.com/scrapy/scrapyd/issues/246

when scrapyd-deploy ,can auto call pip install .


Digenis:
Hi @cleocn,
Thank you for your contribution.
Unfortunately, this is not the way we want to fix it.
There should be some separation between projects' environments.
I think scrapyd is not ready for such a change.

cleocn:
agree to you. maybe something like pyenv/virtualenv it's a good choice.
and thank you for the great scrapyd.

but it can save a lot of time for me .
because i run  scrapyd in docker.
installed packages will lost when i restart the docker container.

GrazianoFracasso:
Hi all, i run scrapyd on docker, I already read that you don't want to resolve this way, but I forked cleocn:master in order to resolve pip main error. ( https://stackoverflow.com/questions/49839610/attributeerror-module-pip-has-no-attribute-main/49990889 pip main is not supported in latest versions ).
[https://github.com/GrazianoFracasso/scrapyd/commit/a4d83aa1ed6fcdf8c86610c5c8656b44f55a156f](a4d83aa1ed6fcdf8c86610c5c8656b44f55a156f)

georgescumihai:
Ok, so what is the workaround to install requirements ?

jpmckinney:
Closing per above comment by maintainer. Related: #224 and https://github.com/scrapy/scrapyd-client/issues/79

PR #272, opened 2018-01-26
Pass scrapy arguments to ScrapyProcessProtocol 

dfrank-a:
Pass scrapy arguments along to ScrapyProcessProtocol.

This will allow users to use this data in extensions of the minimal admin UI in whatever way they see fit.

### DIFF
--- a/scrapyd/launcher.py
+++ b/scrapyd/launcher.py
@@ -42,7 +42,7 @@ class Launcher(Service):
         env = e.get_environment(msg, slot)
         env = native_stringify_dict(env, keys_only=False)
         pp = ScrapyProcessProtocol(slot, project, msg['_spider'], \
-            msg['_job'], env)
+            msg['_job'], env, args)
         pp.deferred.addBoth(self._process_finished, slot)
         reactor.spawnProcess(pp, sys.executable, args=args, env=env)
         self.processes[slot] = pp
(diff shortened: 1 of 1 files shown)


Digenis:
Hi @dfrank-a
welcome to scrapyd.

This feature is something we want to add eventually.
I only didn't open an issue
because of worries I express in the following 2 comments of issue 123
(sorry for the mess, I didn't have the time to write an issue)
https://github.com/scrapy/scrapyd/pull/123#issuecomment-179128857
https://github.com/scrapy/scrapyd/pull/123#issuecomment-180403879
To sum them up, one day we may need to write a new webservice
to replace schedule.json with one that nests spider arguments & settings
in separate objects, so that their names don't collide with special arguments
for the webservice or other scrapyd components.
Will merging this PR make it harder to replace the webservice later?
I really haven't thought much about it and don't have the time now.
What's your opinion on this?

Digenis:
Actually I just took a look and I think it won't make it harder.
This however only means us. It won't make it harder for us.
If users start writing custom components that use the process protocol's args
their components will break when we reorganize the object into something more nested.
Perhaps we should include it as an experimental change
with a warning about the structure of the args attribute.

dfrank-a:
That is something I hadn't considered. How do you feel about declaring a public interface to the data with property methods? This should allow for flexibility in our implementation while maintaining consistency in the external interface.

Digenis:
I see what you mean, it makes sense
but I confused the spider arguments with the scrapy process's args.
Since it's process's args you are adding to the proc. protocol,
the compatibility concerns don't apply here.
To avoid confusion in the future (e.g. if we decide to add the spider arguments)
perhaps args should be named `cmdline`
both in _spawn_process and ProcessProtocol.

Also, the commit message should fit in a single line up to 50 characters.

jpmckinney:
See also #343 about adding priority.

Note: https://github.com/scrapy/scrapyd/pull/343#issuecomment-505096894

> Anyone using a custom laucher, poller, spiderqueue, sqlitequeue will have their code broken.
> It's been long since the last release. Let's make the transition smooth.

i.e. make this a minor version change.

jpmckinney:
Re-done in #474.

PR #275, opened 2018-03-14
new_href

torm89:
New href allows run website with more complex domain address.


Digenis:
Hi @tomaszrycz,
can you give me an example where this is a problem?
Do you use something like apache's mod_userdir?

jpmckinney:
Closing as no response to maintainer's question after 2 years.

jpmckinney:
Fixed by #482

Issue #234, opened 2017-06-19
Starting scrapyd docker container with eggs included


VanDavv:
Hi
I've been experimenting a little with scrapyd on docker, and done the following:
- in the config file, i specified different directory for eggs
`eggs_dir          = /src/eggs`
- in dockerfile, i added prebuilt projects to this directory
`ADD eggs /src/eggs`

At first glance, it looked like it's working
![image](https://user-images.githubusercontent.com/5244214/27278365-70fc0730-54e1-11e7-95b2-22ef8c00df35.png)

but, when I wanted to make a scheduje.json post, it returned me an error

> {"node_name": "295e305bea8e", "status": "error", "message": "Scrapy 1.4.0 - no active project\n\nUnknown command: list\n\nUse \"scrapy\" to see available commands\n"}

I could type anything into project and spider fields and the result was the same. 
How can I fix this issue?

amarynets:
Hi, this message means as far as I remember that Scrapy project is not deploying. I'm not working with Docker, but could you just run command scrapyd in a container if it will help, this means that you have a problem with permission to file system. Also sometimes(I don't know why. I have this only on my local machine) if scrapyd service was stopped and restart, the project can display as active in 127.0.0.1:6800 port, but when you will try to run, you can get this error

VanDavv:
I think it's rather because scrapyd, when making request to `addversion`, despite adding egg file to `eggs_dir`, is making some other stuff that activate the project. I have even seen this functions in code, but I'm not able to recreate them. Also, I tried searching in sqlite if there are some data about eggs, but unfortunatelly there wasn't and I'm stuck

amarynets:
I'd recommended you use scrapyd-client for deploy and after deploy run scrapyd server

VanDavv:
scrapyd-client is good on small scale. I wan't to have docker image with eggs and daemon configurated, so that I can launch it right away, without using scrapyd-client or scrapyd API

Digenis:
@VanDavv, the project name in "available project"
shouldn't have the version, python version and egg extension in the name.
There's definitely something wrong there.
We need more info.
You configuration files, the commands you type and their output and the logs.
Eventually, the docker image but unfortunately I don't have the time to dig that far right now.

VanDavv:
Dockerfile
```
FROM python:3.6
MAINTAINER lpilatowski@teonite.com

RUN set -xe \
    && apt-get update \
    && apt-get install -y autoconf \
                          build-essential \
[...]

PR #283, opened 2018-05-16
Add docker image

ruairif:


jpmckinney:
You are welcome to maintain a Docker image independently.

To keep maintenance burden low, we don't include such packaging.

jpmckinney:
Done in ee4e036ec13f73ea2bb58f03c601ef71a5ce1d16. Related: #234

PR #287, opened 2018-05-23
Handle html a tag path correctly

yidasanqian:


Digenis:
I dodn't see how they were wrong.
Also, why only the logs and not the items?

However you have a point.
The "Go back" anchor is wrong.
It should have been called "Go up".
For "Go back", browsers have such a button.

jpmckinney:
Also replaced by #482

PR #292, opened 2018-08-01
update website.py - Cancel column visibility

bsekiewicz:
In the docker environment (python: 3.6-slim) it does not display the Cancel column in the Jobs subpage.


Digenis:
fixed in #315

Issue #199, opened 2017-01-05
Job Status Notification.


monalisir:
Is there any way to notify client which job has been finished ?
So the client do not need to add a timer and check the status every period time.

Digenis:
There is ticket for this, #55, which is low on the priority list
until more users show interest andor someone comes up with a nice PR.

I see two alternatives.
You can either write an extension
and use it in all your deployed projects,
which should be simple for 1-2 projects,
or write a custom [launcher](https://github.com/scrapy/scrapyd/blob/1.1/scrapyd/launcher.py)
and [configure](https://github.com/scrapy/scrapyd/blob/1.1/scrapyd/default_scrapyd.conf#L15) your scrapyd to use it.

my8100:
The third alternatives would be [ScrapydWeb](https://github.com/my8100/scrapydweb), and get email notified when a job is finished.

jeromecaisip:
@Digenis If you don't mind. How are you able to add a path on the scrapyd.conf so I can use my launcher class on my module. Thanks.

Digenis:
Hi @jeromecaisip,

Make a package for your launcher module with setuptools
https://setuptools.readthedocs.io/en/latest/setuptools.html

Install it in the python environment in which scrapyd runs
 eg `pip install git+file:///home/jerome/proprietary-scrapyd-plugins/`

Define the path in your scrapyd configuration.
```
[scrapyd]
launcher = proprietary_scrapyd_plugins.launcher.Launcher
```

This seems to be common question.
Consider contributing a tutorial to the mailing list or a wiki.

jpmckinney:
Closing as duplicate of #55.

PR #300, opened 2018-11-16
add/update/deploy scrapy code from git

mouday:
通过git ssh连接，添加、更新、部署 scrapy项目代码

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ docs/_build
 venv
 dist
 build
+.idea/
+scrapyd/scripts/twistd.pid
+scrapyd/scripts/projects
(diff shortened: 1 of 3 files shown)


Digenis:
Hi @mouday,
I understand what you want to do
but there are many reasons why I can't merge this PR.
Even if you fix the code and document it
I'm not convinced that it's in the scope of scrapyd to pull & deploy projects from git.

You can actually do this in a custom webservice.
You modified 3 files but all you need is writing a custom webservice.
See my following comments on how to do it.
https://github.com/scrapy/scrapyd/issues/307#issuecomment-483764946
https://github.com/scrapy/scrapyd/issues/199#issuecomment-483775572

Issue #312, opened 2019-04-15
The 'Cancel' button is not visible in Python 3.x


my8100:
It should be ```if b'cancel.json' not in self.root.children:``` for Python 3.x

https://github.com/scrapy/scrapyd/blob/d3f47f3d252373fb34d6fa99a93432909cf72fc9/scrapyd/website.py#L141-L143

Digenis:
I think the real problem is that we read the configuration without regard to encoding.
https://github.com/scrapy/scrapyd/blob/d3f47f3d252373fb34d6fa99a93432909cf72fc9/scrapyd/config.py#L19-L23
Does this work for you?
```python
    self.cp = SafeConfigParser()
    self.cp.readfp(StringIO(default_config))
    self.cp.read(sources)
    for fname in extra_sources:
        with io.open(fname) as fp:
            self.cp.readfp(fp)
```

my8100:
No, it still doesn't work for Python 3
https://github.com/scrapy/scrapyd/blob/d3f47f3d252373fb34d6fa99a93432909cf72fc9/scrapyd/website.py#L141
The result of `print(self.root.children.keys())`:
`dict_keys([b'', b'logs', b'jobs', b'schedule.json', b'cancel.json'...`

my8100:
@Digenis Actually, I really doubt who would deactivate cancel.json.

my8100:
@Digenis
The problem still exists in Python 3.7.
Did you test before merging #315?

Digenis:
fixed in e3c9f91

Digenis:
I only tested whether it reads correctly a file with non-ascii-encodable resource names.
But these were different bugs. Something that I completely forgot when I opened #315.

PR #321, opened 2019-04-20
Add 'Schedule' buttons in the Jobs page

my8100:
This PR adds a schedule endpoint based on #256.
Note that the 'Cancel' and the 'Schedule' buttons still not visible
in Python 3 due to #312.

### DIFF
--- a/scrapyd/website.py
+++ b/scrapyd/website.py
@@ -118,7 +118,15 @@ class Jobs(resource.Resource):
     <form method="post" action="/cancel.json">
     <input type="hidden" name="project" value="{project}"/>
     <input type="hidden" name="job" value="{jobid}"/>
-    <input type="submit" style="float: left;" value="Cancel"/>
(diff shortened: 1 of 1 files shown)


my8100:
@Digenis, What do you think about the added 'Schedule' button?

Digenis:
There's no way to add arguments and settings
and, taken from the finished jobs,
it creates the confusion that the job will run with the same arguments.

The reason I wrote the cancel shortcut is that the solution suits any user.

If we add a form to add arguments, we are getting in to the scope of other tools
such as scrapyweb.

my8100:
Not every spider run needs arguments and settings from cmd.
And it's convenient for repeating crawling jobs with just a click.

PS: typo for *scrapydweb*

jpmckinney:
I share the concerns of @Digenis in https://github.com/scrapy/scrapyd/pull/321#issuecomment-485448325

PR #325, opened 2019-04-24
Listen on unix socket

Digenis:
This allows securing the app from users on the same system,
provided that the effective group id of scrapyd's process is properly chosen.

### DIFF
--- a/scrapyd/app.py
+++ b/scrapyd/app.py
@@ -1,5 +1,7 @@
+import os
+
 from twisted.application.service import Application
-from twisted.application.internet import TimerService, TCPServer
+from twisted.application.internet import TimerService, TCPServer, UNIXServer
 from twisted.web import server
 from twisted.python import log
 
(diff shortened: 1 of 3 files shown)


Digenis:
Perhaps the app should also stop using tcp by default.

pawelmhm:
integrated with current master here https://github.com/scrapy/scrapyd/compare/uds?expand=1

scrapyd/app.py
```diff
@@ -33,15 +38,22 @@ def application(config):
...
+    if http_port:
+        webservice = TCPServer(http_port, website, interface=bind_address)
+        log.msg(format="Scrapyd web console available at http://%(bind_address)s:%(http_port)s/",
+                http_port=http_port, bind_address=bind_address)
+        webservice.setServiceParent(app)
+    if uds_path:
+        webservice = UNIXServer(uds_path, website, mode=0o660)
+        log.msg(format=u"Scrapyd web console available at http+unix://%(uds_path)s",
+                uds_path=uds_path)
+        webservice.setServiceParent(app)
```
pawelmhm, line 57: I'm not certain if I understand the logic here, is this really meant to start both TCPServer and UnixServer? Or is there some intention to allow to disable and stop TCPServer by setting http_port to zero?

Because with this implementation if someone just adds uds_path without removing http_port it will start both of them. Is that expected?

jpmckinney:
This seems to be equally solved by basic authentication, which is already merged. (The Unix socket strategy is to use filesystem permissions to prevent other users from connecting to the socket.)

Since this is not related to any feature request issue (i.e. no clear demand), I'll close.

jpmckinney:
Done in #510

PR #326, opened 2019-04-27
Support basic authentication

my8100:
This PR introduces basic authentication, which would be enabled
if both `username` and `password` options are set up.


my8100:
@Digenis, how do you like this implementation?

Digenis:
I'm very positive towards this PR.
I'll review it more closely soon
and merge it before #325 to avoid bugging you with merge-conflicts.

Btw, thank you for all your contributions
and welcome to the scrapyd project.

my8100:
@Digenis, how is it going?

my8100:
@Digenis how is it going?

scrapyd/default_scrapyd.conf
```diff
@@ -10,6 +10,8 @@ finished_to_keep = 100
 poll_interval = 5.0
 bind_address = 127.0.0.1
 http_port   = 6800
+username    =
+password    =
```
Digenis: I think we can use a single config option
`http_auth = user:pass`
No worries for colon character in username
according to [rfc 7617](https://tools.ietf.org/html/rfc7617#section-2)
> Furthermore, a user-id containing a colon character is invalid, [...]
User-ids containing colons cannot be encoded in user-pass strings

Digenis:
There are many things that need to be considered in this idea.
We should be using digest auth by default,
hashed password, file shouldn't be readable by other users. etc

Overall, I think this PR will add a false sense of security
to users that would otherwise make effort
to put scrapyd behind a vpn or an https proxy that adds its own auth layer.

I will still merge this because for many development scenarios
it's simpler than setting up any of the above.

There should be an unsettling warning about it in the docs.
In production the only inet address it may listen on is locahost, using a password.

I'll add more comments tomorrow.

scrapyd/default_scrapyd.conf
my8100: 1. The `http_auth` option might be confusing as the configuration file of Scrapyd is not self-documented.
2. We should check the `username` option to avoid potential mistake (see [commit 8664360](https://github.com/scrapy/scrapyd/pull/326/commits/866436005f09a2df1b99b0a53e4e67a5e46bf0a6)), just like [htpasswd](https://httpd.apache.org/docs/2.4/programs/htpasswd.html):
```
$ htpasswd .htpassfile user:name
htpasswd: username contains illegal character ':'
```

Digenis:
@my8100, are you ready?
shall I merge?

my8100:
@Digenis
Yes, go ahead!

Digenis:
@my8100,
What do you think of moving all the warping code from the app to the website module?

my8100:
> @my8100,
> What do you think of moving all the warping code from the app to the website module?

OK, l would add a new function named get_resource and send a PR.

my8100:
@Digenis 
See PR #337

Digenis:
I was actually asking what do you think of the idea.

My thought is that we can override its constructor
to return the instance wrapped in the other classes
if the config defines credentials.

A constructor returning an instance of a different class
is probably better written as a factory,
just like as you did,
but since here we're using interfaces,
the fact that the wrapper class implements
all the interfaces implemented by the webclass
make it quite acceptable, don't you think?

my8100:
I thought you simply want to slim the app module.

I think PR #339 is making things complicated,
especially for those want to override the Root class.
It's better to keep the auth layer transparent to the webcls.

Digenis:
It depends on what they want to override.
People who want to write a custom authentication layer
need to remember to come up with a different config option
for user/pass because they are used by the app.
I think whatever configuration concerns http, like authorization,
should be he handled by the website.

I think you are actually concerned more about users who want to override the webcls
while keeping the auth layer, without copy/pasting code from the built in website.
That's a valid concern.
The code that wraps a resource in an auth layer
can be put in a function called by `__new__`
so that users defining custom websites can reuse it.
https://github.com/scrapy/scrapyd/pull/339/commits/28e2934bee6e4ee6408bd7e12fdac4e38c6a89bb

PR #328, opened 2019-04-29
Avoid using DeferredQueue.pending to fix #327

### LINKED ISSUE #327: The job in poller.dq.pending is not shown as a pending job when reaching max_proc
There could be one job stored in the list of `self.root.poller.dq.pending` when reaching max_proc,
like `[{'settings': {}, '_job': '5', '_project': 'ScrapydWeb_demo', '_spider': 'test'}]`,
which would be the first job to run when the amount of running jobs is less than `max_proc`
https://github.com/scrapy/scrapyd/blob/21135a014c839d21cc0157500b44a6359c807c1d/scrapyd/poller.py#L14

https://github.com/twisted/twisted/blob/59f8266c286e2b073ddb05c70317ac20693f2b0c/src/twisted/internet/defer.py#L1865-L1877
```
class DeferredQueue(object):

    def put(self, obj):
        """
        Add an object to this queue.

        @raise QueueOverflow: Too many objects are in this queue.
        """
        if self.waiting:
            self.waiting.pop(0).callback(obj)
        elif self.size is None or len(self.pending) < self.size:
            self.pending.append(obj)
        else:
            raise QueueOverflow()

```

But Scrapyd only uses `self.root.poller.queues` when handling pending jobs,
such a problem also exists in the API of `cancel.json`, `listprojects.json`, and `daemonstatus.json`.

https://github.com/scrapy/scrapyd/blob/21135a014c839d21cc0157500b44a6359c807c1d/scrapyd/website.py#L187-L195

### Steps to reproduce the problem:
0. Set `max_proc_per_cpu = 1` 
`[Launcher] Scrapyd 1.2.0 started: max_proc=4, runner='scrapyd.runner'`

1. Launch 10 jobs in total
image

2. Jobs No.1 to No.4 are running, **but job No.5 is not shown as a pending job**
image

3. Manually cancel job No.1, then job No.5 starts to run, **but job No.6 is not shown as a pending job**
image

4. Finally, all 10 jobs are shown as finished jobs
image

my8100:
This PR keeps `poller.dq.pending` empty so that all pending jobs
would be shown correctly in the Jobs page, as well as the API of
`listjobs.json` and `daemonstatus.json`.

### DIFF
--- a/scrapyd/poller.py
+++ b/scrapyd/poller.py
@@ -11,11 +11,11 @@ class QueuePoller(object):
     def __init__(self, config):
         self.config = config
         self.update_projects()
-        self.dq = DeferredQueue(size=1)
+        self.dq = DeferredQueue()
 
     @inlineCallbacks
     def poll(self):
-        if self.dq.pending:
+        if not self.dq.waiting:
             return
         for p, q in iteritems(self.queues):
             c = yield maybeDeferred(q.count)


Digenis:
So, if I understand correctly,
the dq was a redundant queue after our sqlite queue, right?

my8100:
No, because the `Launcher` need to control `max_proc` by using `poller.dq.waiting`. 
But there is no need to put one job in the `poller.dq.pending` when reaching `max_proc`.

my8100:
The implementation of the `put` method of `DeferredQueue`:
https://github.com/twisted/twisted/blob/59f8266c286e2b073ddb05c70317ac20693f2b0c/src/twisted/internet/defer.py#L1865-L1877

jpmckinney:
This solution works if there is a single project. However, if there are two projects (for example), then once one crawl finishes, the poller will pop the "dummy" item in the `waiting` list and fire the new crawl for a first project, but then it will append a crawl for the second project to the `pending` list. So, the `waiting` check needs to be performed on each iteration.

PR #330, opened 2019-05-03
Add api to get the job's stats

Tzeross:
Parse the log and get the stats you output in scrapy
image


Tzeross:
oh

my8100:
FYI: https://github.com/my8100/logparser

jpmckinney:
Looks like author intended to push this branch to their own repo, so marking as invalid.

This use case can generally be satisfied by accessing the log file via Scrapyd.

PR #337, opened 2019-06-16
Add function get_resource in the app module 

my8100:
The get_resource func keeps the auth layer transparent to the webcls.

### DIFF
--- a/scrapyd/app.py
+++ b/scrapyd/app.py
@@ -16,7 +16,7 @@ from .environ import Environment
 from .basicauth import PublicHTMLRealm, StringCredentialsChecker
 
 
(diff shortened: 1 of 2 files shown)


Digenis:
Either the first commit or a squash of both commits is fine.
I'm inclined towards the first.

If you squash them, don't forget to change the message
to something like "move auth wrapping code to function"
instead of "move get_resource ...".

Factory functions are usually prefixed with "create"
but here its purpose is restricted to wrapping the website.
Maybe create_wrapped_resource if it's not too verbose.
Your thoughts/ideas?
We can also just merge it as it is.

PR #346, opened 2019-06-25
Avoid cancelling running jobs of all projects

### LINKED ISSUE #345: cancel.json terminates running jobs of all projects
Both running jobs ('project1', 'jobid') and  ('project2', 'jobid')
would be terminated when cancelling ('project1', 'jobid').

my8100:
Fix #345.

### DIFF
--- a/scrapyd/webservice.py
+++ b/scrapyd/webservice.py
@@ -72,7 +72,7 @@ class Cancel(WsResource):
             prevstate = "pending"
         spiders = self.root.launcher.processes.values()
         for s in spiders:
-            if s.job == jobid:
+            if s.project == project and s.job == jobid:
                 s.transport.signalProcess(signal)
                 prevstate = "running"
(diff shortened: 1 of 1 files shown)


Digenis:
The commit message can be a bit confusing for someone who doesn't know about the bug.
Soft words such as "avoid" can make the sentence read like it's just a slight change in behaviour.

In commit messages for bugfixes,
I usually find in past tenses an easy way to simultaneously
describe the old behaviour, make it easy to assume what changed and imply that it was a bug,
in a single sentence.

e.g. Cancel ws was killing running jobs by jobid alone.

Digenis:
1.2.2 cherry-pick candidate

my8100:
@Digenis 
Thank you so much for sharing your experience.
For me, sometimes writing a proper and precise commit message is much harder than fixing a bug.
: )

PR #347, opened 2019-07-07
Add an option to remove pending jobs at startup

my8100:
`remove_pending_jobs = off`

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -109,6 +109,15 @@ This refers to logs and items.
 
 This setting was named ``logs_to_keep`` in previous versions.
 
+.. _remove_pending_jobs:
+
+remove_pending_jobs
+----------------
+
+.. versionadded:: 1.3
+
+Whether to remove pending jobs at startup. Defaults to ``off``.
+
 .. _finished_to_keep:
 
 finished_to_keep
--- a/scrapyd/default_scrapyd.conf
+++ b/scrapyd/default_scrapyd.conf
@@ -6,6 +6,7 @@ jobs_to_keep = 5
 dbs_dir     = dbs
 max_proc    = 0
 max_proc_per_cpu = 4
+remove_pending_jobs = off
 finished_to_keep = 100
 poll_interval = 5.0
 bind_address = 127.0.0.1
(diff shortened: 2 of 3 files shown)


my8100:
@Digenis
How about this PR?

Digenis:
Are you sure this is not something that cannot be accomplished by the init/systemd script?
E.g. putting the databases in a /run/ directory which is cleaned up before running.
I think deciding to discard scheduled runs is a responsibility of whatever scheduled them.
However, I see usefulness in this, eg when recovering from a period of service unavailability.
Instead of accumulating multiple pending jobs with adjacent crawling scopes¹
users may implement the downtime-compensating extension of the crawling scope
in the spider itself. The web is messy, you often can't separate the logic of planing the crawl
and performing it. Do you have other use-cases for motivation?


Footnote
¹ Sorry for the terminology, I had to make up some terms that may be unidiomatic.
By crawling scope I mean definable limits in a spider's domain (not domain name).
E.g. Crawling for price comparison: the scope may be limited by product categories.
Crawling blogs: the scope may be limited to the last 24h published material
and the adjacent scope for which the spider will compensate is from 48 to 24 hours ago.

my8100:
1. remove_pending_jobs() is executed after update_projects() in class QueuePoller,
so there's no need to worry about the dbs_dir.

2. Some users may just want to remove all pending jobs by restarting Scrapyd.
3. I think it would be useful if we can **provide command-line interface**,
e.g.  --remove_pending_jobs, --stop_running_jobs, and --list_config.
But it seems that we have to [write a twistd plugin](https://twistedmatrix.com/documents/current/core/howto/tap.html) to add subcommands to the twistd command.

Digenis:
Perhaps the twisted plugin is unavoidable because of #70.

Again, are you sure you need this?
In the current state of scrapyd
it takes only a line or two in an init script.

If you want to set a promise
that later or customized versions of the poller
that may use something more complex that sqlite files
must provide this method too
you'll want to add it to the poller's interface.

jpmckinney:
Closing as not related to a feature request issue.

PR is missing an update to the IPoller interface.

Also, as Digenis mentions, for the default queue class, you can instead just delete the files in the `dbs_dir`.

PR #348, opened 2019-07-13
Log useful messages when reading configuration

my8100:


Digenis:
Some things in the config are leftovers that nobody got around to delete them.
e.g.  There's no straightforward way to set the extra sources
unless you are running scrapyd from a scrapy command.
This script hasn't been updated since 2013, I believe no one is using it.

I think you are trying to to solve the problem that a user has to make extra effort
to figure out what is the effective configuration and where it was red from.
But why would you log, say, the log path but not the item path?
Why the config path but not the dbs dir?
We can go on logging all possible config options used by scrapyd
but not only would this be noise in the log
it would be noise in the code too, full of if/else statements.
I would consider logging (in debug level) the whole effective config
produced by configparser.write(StringIO).

jpmckinney:
I am also not sure what problem (e.g. issue) this is resolving.

jpmckinney:
Per above comments, if we were to do this, we should either log the config file name only, or also log the effective configuration.

I understand that this could be useful for debugging. That said, in my own deployments, I've only ever had the need for a single configuration file (whichever you prefer).

(I don't really understand the use case for having so many potential paths for configuration.)

PR #349, opened 2019-07-18
Fix polling order in contrib modules

my8100:
This PR use project_priority_map to store (priority, -timestamp) as value,
so as to respect both job priorities across projects and the FIFO principle.

See also issue #187 and PR #344.

### DIFF
--- a/scrapyd/app.py
+++ b/scrapyd/app.py
@@ -13,6 +13,7 @@ from .interfaces import IEggStorage, IPoller, ISpiderScheduler, IEnvironment
 from .eggstorage import FilesystemEggStorage
 from .scheduler import SpiderScheduler
 from .poller import QueuePoller
+from .contrib.fix_poll_order.poller import FixQueuePoller
 from .environ import Environment
 from .config import Config
 from .basicauth import PublicHTMLRealm, StringCredentialsChecker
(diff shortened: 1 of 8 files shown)


my8100:
@Digenis
How about this contrib implementation?

Digenis:
I was busy, I'll review it in a day or two

my8100:
Any progress?

my8100:
Any progress?

scrapyd/app.py
```diff
@@ -23,7 +24,11 @@ def application(config):
     bind_address = config.get('bind_address', '127.0.0.1')
     poll_interval = config.getfloat('poll_interval', 5)
 
-    poller = QueuePoller(config)
+    if config.getboolean('fix_poll_order', False):
+        log.msg("Activating contrib modules for fix_poll_order")
+        poller = FixQueuePoller(config)
+    else:
+        poller = QueuePoller(config)
```
jpmckinney, line 31: Should do something similar to Launcher, etc. so that users can select any module (including their own), not just between these two modules.

jpmckinney:
This looks similar to #344 except is puts the new logic in a contrib module, and users opt-in to using that module.

As mentioned there, this multi-queue prioritization approach seems more complicated than a single queue, discussed in #187. I've now made the `poller` configurable, so a user *could* copy this code into their own project, and load it, if they prefer this approach.

PR #409, opened 2021-09-30
Replaced FEED_URI with FEEDS argument

msenior85:
Replaced FEED_URI spider argument with the FEEDS argument as the FEED_URI argument is deprecated in scrapy version 2.1.0


jpmckinney:
Replaced by #471

Issue #117, opened 2015-11-11
Project name sanitization


csalazar:
As described [here](http://www.spect.cl/blog/2015/11/security-audit-scrapyd/) (point 1,2,3 and 5) there are some security issues related to **project name**. I've thought in a fix sanitizing the project name value using the same logic as with variable `version`:

```
re.sub(r'[^a-zA-Z0-9_-]', '_', version)
```

It would have to apply to every method working with project name in `FilesystemEggStorage` to have consistency adding and then getting projects. It adds as side efect that two projects with non-common characters like `project!` and `project?` will share the same project name `project_`. Does someone see a better solution?

Digenis:
Percentage encoding them should also prevent name collisions.
This should also take care of the XSS issues to a point.
Still if someone succeeds to store an unescaped egg
the site module should ignore it and maybe warn about it.

One thing to note however before any development on this bug.
**project name != project package**
I can rename my default project package to something shorter
to keep its imports short
or the other way around, rename the project name to something longer
to make its name more descriptive than the package.
I do it all the time by modifing `scrapy.cfg`.

Last, I don't think these are security vulnerabilities.
Scrapyd is overly permissive and non-isolating by design.
However I agree that all these are bugs and need to be fixed
as I also believe it needs a total redesign.
The fixes probably break backwards compatibility
but mostly in scenarios that the bugs where used as features.

I remember "scrapyd2" being referenced by one of the maintainers
but I think they will not release such a package,
because they the work on the cloud platform.
They gave maintenance to @jayzeng 
which pretty much means that the ball is on the community's court.

csalazar:
Hi @Digenis , I don't consider using percentage encoding the right way. When we insert the project name at root page, we have to display as it is (`<img onerror=alert(1); src=x>`) but taking care of special characters, that's HTML encoding and percentage (URL) encoding doesn't solve anything at that context. 

I would prefer an intermediary representation like `base64` which doesn't generate characters with meaning at filesystem level (slash, dots, etc) so we can deal with project name safely and when we have to display it, we choose the right encoding according to the context.

I think that they are vulnerabilities since my scenario to do the audit was thinking about someone using *scrapyd* and running the spiders in containers, so the issues related to python code execution are restricted at container level and he could use *scrapyd* "safely".

Digenis:
The base64 alphabet has "/"
and also makes the project names unreadable.
I can only think of one pitfall in the %-encoding
[...]

PR #421, opened 2021-11-29
Sanitize project name

pawelmhm:
This fixes important security problem https://github.com/scrapy/scrapyd/issues/117 and also adds unit tests for webservice, increases coverage to 80%.

### DIFF
--- a/scrapyd/webservice.py
+++ b/scrapyd/webservice.py
@@ -16,7 +16,7 @@ from .utils import get_spider_list, JsonResource, UtilsCache, \
 def with_safe_project_name(func):
     @functools.wraps(func)
     def wrapper(resource, txrequest):
-        project_name = txrequest.args.get(b'project', [None])[0]
+        project_name = txrequest.args.pop(b'project', [None])[0]
         msg = "'Project' name is required and must not contain illegal characters. "
         msg += f'Project name "{project_name}" is not valid'
         if not project_name:
@@ -65,12 +65,12 @@ class DaemonStatus(WsResource):
 
 class Schedule(WsResource):
 
-    def render_POST(self, txrequest):
+    @with_safe_project_name
+    def render_POST(self, txrequest, project):
         args = native_stringify_dict(copy(txrequest.args), keys_only=False)
         settings = args.pop('setting', [])
         settings = dict(x.split('=', 1) for x in settings)
         args = dict((k, v[0]) for k, v in args.items())
-        project = args.pop('project')
         spider = args.pop('spider')
         version = args.get('_version', '')
         priority = float(args.pop('priority', 0))
--- a/scrapyd/webservice.py
+++ b/scrapyd/webservice.py
@@ -17,7 +17,7 @@ def with_safe_project_name(func):
     @functools.wraps(func)
     def wrapper(resource, txrequest):
         project_name = txrequest.args.get(b'project', [None])[0]
-        msg = "'Project' name is required and must be valid name allowed by setup.py"
+        msg = "'Project' name is required and must not contain illegal characters. "
         msg += f'Project name "{project_name}" is not valid'
         if not project_name:
             raise Error(code=400, message=msg.encode())
@@ -126,8 +126,8 @@ class AddVersion(WsResource):
 
 
 class ListProjects(WsResource):
-
-    def render_GET(self, txrequest):
+    @with_safe_project_name
+    def render_GET(self, txrequest, project):
         projects = list(self.root.scheduler.list_projects())
         return {"node_name": self.root.nodename, "status": "ok",
                 "projects": projects}
--- a/scrapyd/webservice.py
+++ b/scrapyd/webservice.py
@@ -1,6 +1,11 @@
+import re
 from copy import copy
 import traceback
 import uuid
+
+from twisted.web.error import Error
+from twisted.web.http import Request
+
 try:
     from cStringIO import StringIO as BytesIO
 except ImportError:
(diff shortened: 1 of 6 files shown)


scrapyd/webservice.py
```diff
@@ -1,67 +1,97 @@
...
+    def wrapper(resource, txrequest):
+        project_name = txrequest.args.pop(b'project', [None])[0]
+        msg = "'Project' name is required and must not contain illegal characters. "
+        msg += f'Project name "{project_name}" is not valid'
+        if not project_name:
+            raise Error(code=400, message=msg.encode())
+
+        project_name = project_name.decode()
+
+        allowed_name = re.sub('[^A-Za-z0-9.]+', '-', project_name)
```
jpmckinney, line 27: As @Digenis wrote in #117, this will break backwards compatibility. If I have an existing project name that contains `@` or `_` or `-` or any number of other legal, secure characters, then I can no longer interact with that project using Scrapyd's API. We can maybe make it "opt-in" behavior, and change the default [configuration file](https://scrapyd.readthedocs.io/en/stable/config.html) to opt-in to this behavior. That way, old deployments that don't have the setting in the conf file will not perform this sanitization.

pawelmhm: > we can maybe make it "opt-in" behavior, and change the default configuration file to opt-in to this behavior. 

the goal of this is to secure project from doing something like POST file to /etc/hosts or worse deleteversion /etc/hosts. We should not make this opt-in, should be built in. There is always an option to override scrapyd resources, and adding your handler in scrapyd configuration, you can for example add your own addversion.json. 

> If I have an existing project name that contains @ or _ or - or any number of other legal, secure characters, then I can no longer interact with that project using Scrapyd's API.

yeah you're right, after thinking about this, maybe the approach in this commit is too aggressive, and can break things for some valid use cases. Maybe we could release a milder version in 1.3 and add some deprecation warning saying that in next version, say 2.0. Project names will not accept some characters? Now question which characters to allow in next release, I would say probably all characters that are ok in setuptools.setup(name=) function? So for instance `@` won't be allowed, but space should be allowed? 

To sum up I'll change this to disallow only insecure characters, allow some legal secure characters. Let me know what you think.

pawelmhm: Or maybe let's do it differently, let's add project name sanitization in eggstorage in this PR. Fix project filename there, like we fix version. And in web service just print some deprecation warning in logs about invalid names

jpmckinney: My thinking was to make it opt-in for **existing** projects in 1.x, opt-out (i.e. built-in) for **new** projects in 1.x, and opt-out for **all** projects in 2.x. This could be done by having a configuration setting whose default value is False in 1.x, but True in 2.x. The default template for the configuration file would set the value to True, such that existing projects (whose configuration file doesn't set a value) would use the default value of False, but new projects would have a configuration file that sets it to True.

That said, if we can relax the validation to avoid breaking changes (your first suggestion), or if we can do something else that progressively changes the behavior (if I understand your second suggestion), that's fine too.

pawelmhm: > That said, if we can relax the validation to avoid breaking changes (your first suggestion), or if we can do something else that progressively changes the behavior (if I understand your second suggestion), that's fine too.

I would prefer progressive enhancement, but there is no way to completely avoid breaking it for some people here. If someone was storing projects with unusual characters, it is always begging for trouble. People who have projects "some/project" will have problem. But if anyone is actually doing something like this, it's on their own risk. 

I think we can trust users to have some knowledge and experience. If someone can install and update ScrapyD he should also be able to read documentation, revert to previous version, remove eggs with bad filenames and deploy new ones using API. 

 We will also give warning in docs about it.

pawelmhm: > That said, if we can relax the validation to avoid breaking changes (your first suggestion)

I relaxed validation here https://github.com/scrapy/scrapyd/pull/421/commits/49ab5a77ce5690f5dec1caccfd337c4bd2234c1a, but it is not completely avoiding breaking changes. Still, I think there is no way to avoid it, if someone uses project name "@my project@" he should really prepare for trouble. 

> My thinking was to make it opt-in for existing projects in 1.x, opt-out (i.e. built-in) for new projects in 1.x, and opt-out for all projects in 2.x. This could be done by having a configuration setting whose default value is False in 1.x, but True in 2.x. 

I thought about it again, it is not a bad idea, but it would add complexity to the project and increase development time and future maintenance time. But I thought about something else, how about we just launch 2.0 in 2022 and just announce this and other changes? 2.0 gives some room for breaking changes and people will tolerate them. It won't be a big 2.0 I guess, because nothing big will change, but it would justify making backward incompatible change here.

jpmckinney: 2.0 instead of 1.3 is fine with me! I'm not too precious about the size of releases :)

scrapyd/utils.py
```diff
@@ -154,3 +155,9 @@ def _to_native_str(text, encoding='utf-8', errors='strict'):
         return text.encode(encoding, errors)
     else:
         return text.decode(encoding, errors)
+
+
+def check_disallowed_characters(text):
+    # Anything that is not in this list: A-Za-z0-9.\s_- is banned
+    disallowed_characters_regex = r'[^\"A-Za-z0-9.\s_-]+'
```
jpmckinney, line 162: Just curious - why are whitespace `\s` and the quotation mark `"` included?

jpmckinney (approved):

scrapyd/utils.py
pawelmhm: I admit this is somewhat arbitrary and subjective, but I thought that "project name" is ok name, that someone could make without any mistake. Same for quotes, you can for example by mistake add prroject name with quotes inside scrapy.cfg. 

Creating the list of allowed characters is somewhat arbitrary, we could either go one extreme way and allow only digits and ascii, or go the other way and allow everything except things that are path separators. Path separators should be always disabled so that it stops someone running  scrapyd with sudo user from placing or deleting files anywhere in filesystem. 

I'm open for some feedback and suggestions here, maybe we can also check how other projects are doing things like this.

jpmckinney: The choice makes sense. I'm happy to go with it, as-is. You can maybe add a comment e.g. `Allow the same characters as Python packages, and allow project names like "my project".`

pawelmhm:
Tests unrelated to project name problem merged to master, this Pull Request will only contain sanitization of project name, to be merged later.

jpmckinney (approved):
Approved to squash and merge when working on 2.0.

PR #427, opened 2021-12-24
add publish GH workflow, bump version to 1.3.0

pawelmhm:


pawelmhm:
I'd like to merge https://github.com/scrapy/scrapyd/pull/425 on Monday, and then I think we're good to release 1.3.0. 

There is lots of other work to be done, but it can be done on other versions.

docs/news.rst
```diff
@@ -5,21 +5,28 @@ Release notes
...
+- support for HTTP authentication in scrapyd server
+- replaced deprecated SafeConfigParser with ConfigParser
+- much improved test coverage
+- improved HTTP headers in webservice
 
 Removed
 ~~~~~~~
 
+- Python 2 support
+- Python
```
jpmckinney, line 29: Missing content?

pawelmhm: yes sorry, christmas mood will fix

pawelmhm: I updated news, now just waiting for pypi token and we can release @Gallaecio can you add secret pypi token here?

jpmckinney (approved):

docs/news.rst
Gallaecio: `PYPI_TOKEN` defined.

Gallaecio (approved):

Rotzbua (changes requested):
Just a tiny note:
There is an outdated requirement in the docs `Python 2.7` https://github.com/scrapy/scrapyd/blame/9fae00bb76813359af6d709a343df0f46f529eda/docs/install.rst#L15

pawelmhm:
thanks @Rotzbua I fixed that. Just a note we should also remove six from project, but maybe on a separate PR and with next version.

PR #429, opened 2022-01-12
[setup.py] remove enum-compat

pawelmhm:
it was a requirement for Python 3.4 which we don't need anymore. Also change
required Twisted and Scrapy version to ones that support Python 3.6, versions below
these numbers don't support Python 3.6 which is minimum Python we require now

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -13,7 +13,8 @@ Scrapyd depends on the following libraries, but the installation process
 takes care of installing the missing ones:
 
 * Python 3.6 or above
-* Scrapy 2 or above
+* Scrapy 1.2 or above
+* Twisted 17.9 or above
 
 Installing Scrapyd (generic way)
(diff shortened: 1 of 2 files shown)


jpmckinney:
Why is Scrapy set to >= 1.2? That version is very old.

pawelmhm:
It was set to above 1 before.

We should target Scrapy version that supports Python 3.6, it might be good
to run some tests in github actions with different Scrapy versions by the
way.

I see first version with Python 3.6 support is 1.4 so this here should be
1.4 not 1.2, sorry my mistake.



W dniu śr., 12.01.2022 o 16:18 James McKinney ***@***.***>
napisał(a):

jpmckinney:
Sounds fine to me - but not urgent, as I doubt anyone is trying to run the latest Scrapyd with a very old Scrapy.

PR #431, opened 2022-01-12
[issue-415] Print release version with flag '-v' and '--version'

niuguy:
This is not that elegant as I tried not to mess the original twisted options, if you have a better idea, please reply:)

### DIFF
--- a/scrapyd/scripts/scrapyd_run.py
+++ b/scrapyd/scripts/scrapyd_run.py
@@ -4,11 +4,21 @@ from twisted.scripts.twistd import run
 from os.path import join, dirname
 from sys import argv
 import scrapyd
+import pkgutil
+
+
+def print_version():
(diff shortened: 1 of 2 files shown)


scrapyd/scripts/scrapyd_run.py
```diff
@@ -4,11 +4,21 @@
...
+import pkgutil
+
+
+def print_version():
+    __version__ = pkgutil.get_data(__package__, '../VERSION').decode('ascii').strip()
+    print(f'Scrapyd {__version__}')
 
 
 def main():
+    if len(argv) > 1 and argv[-1] in ['-v', '--version']:
```
jpmckinney, line 16: Hmm, should we just check whether the options appear in any of the argv arguments, e.g.

```suggestion
    if len(argv) > 1 and set(argv[1:]) & set(['-v', '--version']):
```

niuguy: Yes it looks reasonable

PR #434, opened 2022-03-04
[eggstorage] remove deprecated distutils.LooseVersion

### LINKED ISSUE #426: Replace distutils.version.LooseVersion with packaging.version Version
In scrapyd/eggstorage.py we use LooseVersion from distutils https://github.com/scrapy/scrapyd/blob/b45c20547b91f34938ee574f86ba966d4b17b05c/scrapyd/eggstorage.py#L37

The distutils package is deprecated and slated for removal in Python 3.12. 

[PEP 632](https://www.python.org/dev/peps/pep-0632/) recommends using [packaging.version](https://packaging.pypa.io/en/latest/version.html). There is one problem, that LooseVersion was designed [as a tool for anarchists](https://github.com/python/cpython/blob/f4c03484da59049eb62a9bf7777b963e2267d187/Lib/distutils/version.py#L269) and packaging.version is not a tool for anarchists, it raises error when version is not matching [PEP 440](https://www.python.org/dev/peps/pep-0440/). I suspect this may break some projects because many projects don't use semantic versioning but just use version control hashes as version numbers. 

IMO we should just rewrite this sorting in eggstorage to keep old behavior without distutils, I don't see a problem in hashes as version numbers, so we can just keep old way, but maybe worth discussing how to do it? What community expects here? What are your practices and expectations here?

Expected output here:

- [x] patch that replaces LooseVersion with something else
- [x] unit tests that show that current behavior is preserved and does not break people projects, they can go somewhere here: https://github.com/scrapy/scrapyd/blob/b45c20547b91f34938ee574f86ba966d4b17b05c/scrapyd/tests/test_eggstorage.py
or here if we add separate utility for sorting https://github.com/scrapy/scrapyd/blob/b45c20547b91f34938ee574f86ba966d4b17b05c/scrapyd/tests/test_utils.py

pawelmhm:
fixes #426

### DIFF
--- a/reqs/requirements-latest.txt
+++ b/reqs/requirements-latest.txt
@@ -1,2 +1,3 @@
 Scrapy>=1.1
 twisted
+packaging
--- a/scrapyd/eggstorage.py
+++ b/scrapyd/eggstorage.py
@@ -2,9 +2,10 @@ import re
 from glob import glob
 from os import path, makedirs, remove, listdir
 from shutil import copyfileobj, rmtree
-from distutils.version import LooseVersion
 
 from zope.interface import implementer
+
+from scrapyd.utils import sorted_versions
 
 from .interfaces import IEggStorage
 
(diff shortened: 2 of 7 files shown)


pawelmhm:
fixes #434 426

jpmckinney (approved):

jpmckinney:
Redid commits in #469, as easier than resolving conflicts.

PR #439, opened 2022-05-13
Remove unmaintained and unused files

jpmckinney:


jpmckinney:
I'll self-merge this, as these files are unmaintained and unused.

PR #449, opened 2022-06-11
fix: PyPI link for Python Version

Laerte:


Laerte:
@jpmckinney can we merge It?

jpmckinney:
Thanks!

PR #458, opened 2022-11-04
Remove `six` package dependency

Laerte:
We don't need the bindings since we support python 3.6+.

@jpmckinney also: Can i open another PR to [drop support of python 3.6](https://endoflife.date/python) aswell?


jpmckinney:
Yes, can do!

jpmckinney (approved):

PR #459, opened 2022-11-04
Drop Python 3.6 support

Laerte:
@jpmckinney I had added 3.11 support by accident, let me know if you want to add support for this version aswell so i raise other PR too :)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,6 @@ setup(
         'Environment :: No Input/Output (Daemon)',
         'Topic :: Internet :: WWW/HTTP',
     ],
-    python_requires='>=3.7',
     entry_points={
         'console_scripts': [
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -56,6 +55,7 @@
         'Environment :: No Input/Output (Daemon)',
         'Topic :: Internet :: WWW/HTTP',
     ],
+    python_requires='>=3.7',
```
jpmckinney, line 58: I think it's okay to leave this off, since Python 3.6 does work, but we aren't making any guarantees?

jpmckinney:
@Laerte If 3.11 tests pass, we should add it :)

setup.py
Laerte: Sure, this way what will happen is that if someone try to install in a version below 3.7 it will not succeed. Removed.

Issue #475, opened 2023-03-08
SQLite queue is using all CPU on high frequency poller (<1s)


pspsdev:
When running spiders that do nothing at all, the sqlite based poller uses all cpu just reading scheduled tasks. It would be good to have a plug and play alternative queues like redis.

jpmckinney:
Why are you running spiders that "do nothing at all"?

jpmckinney:
Hmm, yeah, same with https://github.com/Tiago-Lira/scrapyd-mongodb (from which scrapyd-redis is forked) and https://github.com/balena/python-pqueue (mentioned in #197).

https://github.com/peter-wangxu/persist-queue is still active, though maybe a first attempt is to switch to https://github.com/scrapy/queuelib as mentioned in #197.

Can you share your setup for reproducing the issue?

jpmckinney:
@pspsdev Now that #476 is merged, do you have suggestions for how to edit the default spider queue, or should there just be a note in the documentation that it doesn't perform well under high frequency polling, and a custom queue would be better?

jpmckinney:
FWIW, I can't replicate this issue. I set `poll_interval = 0.1` and scheduled 100 jobs in a loop.

PR #476, opened 2023-03-08
feat: Add spiderqueue configuration option

### LINKED ISSUE #197: Configurable spider queue class

jpmckinney:
closes #197 

See #475 for updating the default spider queue.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -142,6 +142,16 @@ and services.
 
 For more info see `Twisted Application Framework`_
 
+.. _spiderqueue:
+
+spiderqueue
+-----------
+
+The scheduler enqueues crawls in per-project spider queues, for the poller to pick.
+You can define a custom spider queue class that implements the ISpiderQueue interface.
+
+Defaults to ``scrapyd.spiderqueue.SqliteSpiderQueue``.
+
 .. _webroot:
 
 webroot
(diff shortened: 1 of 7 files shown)


jpmckinney:
Ah, good call, we need to move the `dbs_dir` parsing to the queue's initialization.

For Redis, PostgreSQL, etc. we can have dbs_dir still be a single string, e.g. a connection string like `redis://user:pass@host:port/database` or `postgresql://user:pass@host:port/database`, etc.

jpmckinney:
> Currently the code makes it always a file based solution but scrapyd/spiderqueue.py defaults to :memory: correctly but that can never happen because upstream code always sets a file.

I would say that its use of `:memory:` is incorrect, as it will cause all projects to use the same DB, which is not the intent. The connection string would need to be something like `file:project1?mode=memory` https://www.sqlite.org/inmemorydb.html

jpmckinney:
Okay, custom queues should be more extensible with the latest commit.

PR #471, opened 2023-02-07
Scrapy 2.8 support

jpmckinney:
closes #369
closes #428

PR #500, opened 2024-01-24
DA-243 Upgrade to python 3.12

IonascuDenisa:


jpmckinney:
Replaced by #471. `project_version_exists` logic might be relevant to #517

PR #502, opened 2024-05-24
Added confirm dialog in cancel form

Naik-G:
A confirmation dialog was added to the cancel button to prevent immediate cancellation, as the cancel button is located right after the log, which could lead to unintended cancellations by the user.


jpmckinney (approved):

jpmckinney:
Thank you!

PR #522, opened 2024-07-19
Bisecting Windows CI issues

jpmckinney:


jpmckinney:
It seems it was something about the old Popen wrapper.

PR #540, opened 2025-01-25
Add support for Python 3.13

afuetterer:
Hi @jpmckinney,

this PR adds the Trove classifier and CI test for Python 3.13. scrapyd works fine under 3.13, but after merging it would classified and tested as such.

Let me know what you think.


jpmckinney:
Thank you!

PR #541, opened 2025-01-26
add `requires-python` setting to `pyproject.toml`

afuetterer:
Hi @jpmckinney,

I would like to propose add `requires-python` setting to your `pyproject.toml`. `scrapy` itself already does this:
https://github.com/scrapy/scrapy/blob/c03fb2abb8c354c56c4e8363fc602d49f956c280/pyproject.toml#L51

If you add this setting, `ruff` will also respect that setting, which makes the `target-version` redundant.

Let me know what you think.

Refs:
- https://packaging.python.org/en/latest/specifications/pyproject-toml/#requires-python
- https://docs.astral.sh/ruff/settings/#target-version

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,6 +8,7 @@ version = "1.5.0"
 authors = [{name = "Scrapy developers", email = "info@scrapy.org"}]
 description = "A service for running Scrapy spiders, with an HTTP API"
 readme = "README.rst"
+requires-python = ">=3.9"
 license = {text = "BSD"}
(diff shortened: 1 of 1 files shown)


jpmckinney:
I prefer not to use requires-python, because it causes PyPI and others to not resolve newer package versions when using older Python versions. I don’t officially support older Python versions, but if the package accidentally still works on those versions, I’m happy to let users install the package.

afuetterer:
I see, that is a fair point. Let me close this PR then.

